Issue #70, opened 2016-03-20
Feature: Give users the ability to clone hangouts


lpatmo:
Oftentimes I'll want to schedule a recurring hangout; would be nice if I could click "Copy" and not have to re-fill-out the title and description.

foxiae:
Could we implement a 'reoccuring' feature as well?

Say for example, a channel is going through a book and they meet weekly at x time. So they can just schedule one meeting and have it repeat for x amount of weeks - just like the functionality you have when creating calendar invites.

lpatmo:
@William-R-Wilson Oops, Will, I started working on this bug while I was on the plane earlier today (while I didn't have any wifi) and created this pull request: https://github.com/codebuddiesdotorg/cb-v2-scratch/pull/73. Feedback welcome!

The second part of the bug -- @foxiae's suggestion of giving users the ability to repeat a hangout X number of times weekly from the start -- I haven't tackled, though.

leewaygroups:
So, my understanding is hangout can either be one-off or recurring.

leewaygroups:
Would it make sense if recurring hangout schedule is made more flexible such that the user can pick the days on which the hangout will reoccur?

More like: Time: 12PM - 1PM and days: date1, date2, ... daten.

leewaygroups:
@lpatmo Is this feature (repeating hangouts) still needed?

distalx:
@lpatmo  is this issue has been resolved in PR #73  ?

PR #73, opened 2016-03-21
added clone functionality - solves part of bug #70

lpatmo:


client/templates/hangout/hangout-item.html
```diff
@@ -17,7 +17,9 @@
       </label>
       <ul class="status">
         <li><i class="fa {{getType type}}"></i></li>
-        <li><label>{{getDate this}}</label></li>
+        <li><label>{{getDate this}} |  <a class="clone-hangout">
```
sergeant-q: Logged out users should not be able to either **see** this copy link or we should do the same check we do on a _create new hangout_ to ensure a user is logged in before they can copy a hangout.  Essentially they are creating a new hangout. 

The **check** functionality kicks off a modal asking the user to log in.  See `home.js` starting on `line 12`.

sergeant-q: To add to this. I am receiving the following error when invoking the method `cloneHangout`

**Exception while invoking method 'cloneHangout' Error: Match error: Expected string, got object in field user_id**

This is because there's no user logged in! So the call actually fails and you are not able to copy/create the hangout.

sergeant-q:
I have added some **inline comments** to address some issues. Overall, the functionality is working as expected.  

One thing to note and maybe this is more of a UX issue but when cloning the hangout, it copies over the **start** and **end** times as well.  Maybe we should clear these out and force them to select a new time? I completely overlooked it and when I cloned the hangout, it was already mastered because it happened in the past already!

lpatmo:
Made two changes:

1. Non-logged-in users do not see the "Copy" link at all.
2. When the copyHangoutModal modal pops up, the filled-in start time and end times are the current time and one hour from now, which is what you would expect if you were creating a hangout using the "create hangout" button. So the copying of the title/description/hangout type are the only things "cloned."

Thanks @sergeant-q for the review!

Let me know if you guys see anything else...

Issue #68, opened 2016-03-19
Implement a search box


lpatmo:
Mockup:

![screen shot 2016-03-19 at 12 35 08 am](https://cloud.githubusercontent.com/assets/4512699/13896652/7ed1573c-ed6a-11e5-8eaf-a6b621bb49e1.jpg)

@distalx recommends the meteorhacks:search-source package: https://meteorhacks.com/implementing-an-instant-search-solution-with-meteor.html

distalx:
I'm thinking about rolling out our own custom search, 
Currently I'm considering search 
by `hangout title`
by `tags` (if we use tags)
by `users `

Guys, Please do let me know your thoughts.

sergeant-q:
@distalx I think all of those three search fields are valid options and make sense. Can always just start with those fields and then add additional as we find necessary.

lpatmo:
I would also add `description` to the mix -- since likely there will be some keywords in the description not mentioned in the title. But at a minimum, would be good to search by `hangout title`. 

Thanks for taking this issue, @distalx!

PR #77, opened 2016-03-23
Feature #68 search

distalx:
#68 Search hangout by topic, by description and by user is covered.

### DIFF
--- a/client/css/_home.scss
+++ b/client/css/_home.scss
@@ -33,7 +33,7 @@
         text-decoration: underline;
       }
     }
-    
+
   }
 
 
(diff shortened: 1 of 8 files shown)


lpatmo:
Committed some styling fixes. Great job, @distalx!

sergeant-q:
The search seems to be case sensitive. We will need to fix this.  A search for **ES6** will return but not **es6**. 

@distalx can we fix this?

distalx:
Nice catch @sergeant-q, I'll give a `hotfix` for that in a few minutes.

PR #104, opened 2016-04-10
96 homepage update

lpatmo:


sergeant-q:
The profile image is coming in very pixelated.  I am not sure why.

image

I see we are calling the `userProfileImage` method, which is returning **image_48** from the users document in the db. I see it stores a few variations of the avatar.  Should we maybe use **image_72**?

Lastly, my username is getting wrapped (hence cut-off). Might need to make this container a bit wider?

lpatmo:
Thank you for the review, @sergeant-q!! Will make the fixes soon.

PR #176, opened 2016-07-30
Updated hangout reminder notifications logic

sergeant-q:
Updated logic to ensure hangouts reminder alerts go out for **both** 24 and 2 hour reminder prior to starting for the same hangout.

Note: requirements are to only send email notification for 24 hour reminder and a Slack notification for hangouts starting within **2 hours**.


server/cron_job/methods.js
```diff
@@ -6,7 +6,7 @@
  *
  */
 // if set to true, cron job will run every minute (production default is every 1 hour)
-const debug = false;
+const debug = true;
```
distalx: should be false !?

PR #222, opened 2016-09-28
resolved: learning issue

bizarrejelly5:


server/learnings/publications.js
```diff
@@ -8,9 +8,10 @@ Meteor.publish("learnings", function(limit) {
 
 });
 
-Meteor.publish("ownLearnings", function(limit) {
+Meteor.publish("ownLearnings", function(limit, userId) {
+  check(userId, String);
   if(this.userId){
```
camilleriluke: Is there a reason why you're still using `this.userId` here?

Issue #228, opened 2016-09-28
Inconsistency in hangout's RSVP count.


leewaygroups:
In all hangouts listing page, the RSVP count for a hangout does not match the actual RSVP counts shown in the hangout's detail page.
![joined 1](https://cloud.githubusercontent.com/assets/7301892/18924704/6cd1d6f6-85db-11e6-9a2b-5d8722ceeca0.png)
![joined 2](https://cloud.githubusercontent.com/assets/7301892/18924734/87ee7066-85db-11e6-9df1-4c25e8737513.png)

hi15:
I'm working on it!

adachiu:
@hi15 is @leewaygroups already working on this? 
I'll label this as `in progress`

distalx:
#246 !!

distalx:
@lpatmo here `joined` count  refers to  number of attendees or attendees + host ?

lpatmo:
Number of attendees + host is fine. Merged.

PR #246, opened 2016-09-30
fixed count users(join) in hangout

hi15:
In hangouts listing page, i fix the RSVP count for a hangout to match with actual RSVP counts shown in the hangout's detail page (located in client/templates/status/update-status.html) . And the test is shown below. 
screen shot 2016-09-30 at 3 53 09 pm
screen shot 2016-09-30 at 3 53 24 pm

This is my first pull-request, thanks for opportunity.


distalx:
#228

PR #248, opened 2016-09-30
Issue 182 limit hangouts on profile

William-R-Wilson:
I wasn't sure if I should be using the 'users' array or the 'attendees' array to test against.  Using the Users array right now.


lpatmo:
Thanks @William-R-Wilson and @distalx!

PR #272, opened 2016-10-03
add sort to client side finder for hangouts

William-R-Wilson:
this should be working.  Hangouts were being sorted on the server side, but not on the client, so when a new hangout was created it wasn't sorted until page refresh.


lpatmo:
Thank you @William-R-Wilson!

lpatmo:
Thank you @William-R-Wilson!

PR #276, opened 2016-10-04
Issue 203

Viswajith76:


Briix:
I see some different ways of making the text color white in those commit. Maybe it would be worth it to either change the `$bg-white` variable to just `$white` or create a seperate `$white` variable and use that across the project whenever we want a white text color?

lpatmo:
@Briix If you want to create a new issue and branch + PR with the addition of the new $white variable and some refactoring of $bg-white to $white, feel free!

Issue #275, opened 2016-10-04
Change text on button "Edit Hangout" to "Save Changes" in Edit Hangout Modal


lpatmo:
> The "Edit your Hangout" button can be called "Save changes" instead. Because it's more clear that the changes are being saved. Besides, the modal window already has the "Edit your hangout" title at the very top. From a UX perspective, it would be more clear to the user if the blue button was called "Save changes"
- @adachiu 

![image](https://cloud.githubusercontent.com/assets/4512699/19066617/f6394734-89ce-11e6-85dd-7b7713e7e929.png)

Relevant file to edit: `i18n/en.i18n.json` file.

ollikarppinen:
I can do this one :)

lpatmo:
@ollikarppinen Awesome!

jllamasi:
@lpatmo I took the liberty of doing this issue. There's a PR open, thanks!

lpatmo:
Thanks for the PR, @jllamasi! But working on it when @ollikarppinen had already left a comment to claim it was disrespectful to @ollikarppinen. This isn't a race! Please watch out for the `claimed` label in the future, or coordinate with the person who claimed the issue. Thanks.

PR #278, opened 2016-10-04
changed button text

jllamasi:
Hi there! I just did what issue #275 requested! Hope everything's okay!

### DIFF
--- a/i18n/en.i18n.json
+++ b/i18n/en.i18n.json
@@ -37,6 +37,7 @@
 
 
   // buttons
+  "save_changes": "Save changes",
   "update": "Update",
   "join_hangouts" : "Join Hangouts",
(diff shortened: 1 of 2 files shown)


i18n/en.i18n.json
```diff
@@ -37,6 +37,7 @@
 
 
   // buttons
+  "save_changes": "Save changes",
```
camilleriluke: Since `edit_a_hangout` is not used anymore it can be deleted

jllamasi: @camilleriluke it is still used as modal title

camilleriluke: You are right :) my bad.

lpatmo:
Thanks @jllamasi!

Issue #225, opened 2016-09-28
Instead of hiding the "browse hangouts" behind the logged-in view, make it public


lpatmo:
Proposal:
- [x] 1) Add a "Browse" link to the top of the header menu.
- [x] 2) Allow visitors to click on it to see the logged-in view of the home page.
- [x] 3) If a non-logged-in visitor tries to take an action on the home page, they'll get a popup asking them to log in. 

At the moment our `templates/home/home.html` contains this logic:

```
<div id="home" class="{{#if currentUser}}logged-in{{/if}}">
  {{#unless currentUser}}
      {{> homeLoggedOut}}
  {{else}}
```

We'll have to change logic a bit so that the `home` template and the `home-logged-out` template refer to different URLs. Happy to think over/discuss implementation with another collaborator.

lpatmo:
The advantages to making the logged-in view public, IMO:
- less mystery for new visitors who might wonder how the hangouts + the learnings section work
- more incentive to sign up to participate, maybe?

One note: I'm not sure how the "X users online learning right now" list will show anonymous users; will have to test that out.

adachiu:
@lpatmo Instead of having the "browse" link on header menu to open up a different page, would it be simpler if we replaced the homepage's "join hangouts" screenshot with the logged-in cards view (minus the "Today I Learned" sidebar"). This way new users will get a glimpse immediately when they first arrive to the home page.

Side note: that image is also now outdated since the UI is going to be cards. We should probably replace it anyway to reflect the actual appearance of the hangout cards.

![screen shot 2016-09-28 at 1 50 58 pm](https://cloud.githubusercontent.com/assets/13413986/18931823/a2aa0126-8582-11e6-81c1-47cfb444b85b.png)

lpatmo:
@adachiu Code-wise it might actually take a few more steps to dynamically load in the hangouts here in place of the image, but I'm not against the idea of replacing this image with loaded-in posts. :) 

My concern is whether we want new visitors to see:
- that we *do* have a learnings tab
- that there's a "What did you learn?" area underneath each hangout (this feature isn't implemented yet, but on the roadmap)
- the full history of hangouts that have happened, not just the most recent 6 (since this area is limited in dimension)

What are your thoughts?

adachiu:
[...]

Issue #203, opened 2016-09-18
IDEA: Home(Dashboard) Redesign using CSS Cards


sergeant-q:
I am opening this up for feedback and suggestions but it's an idea. The current design of the two column grid with the hangouts listed on the left hand side seems a bit difficult to consume and discover.

I am proposing a redesign that incorporates a "**card-based**" design. Below is an example for inspiration:

![google now card design](http://thenextweb.com/wp-content/blogs.dir/1/files/2015/06/Screen-Shot-2015-06-16-at-08.57.56.png)
### Another example:

https://dribbble.com/shots/1475597-Mail-Client-UI-Concept/attachments/220136
### Benefits
- Easier to view/discover hangouts
- Appealing and better UX
- Mobile friendly

lpatmo:
Agreed! The home page looks better with a narrower sidebar (8-4 instead of 7-5 grid), and having cards will ensure that a user can see more about what's going on without having to scroll.

I need to mock up where the date should go on each card (maybe at the top).

adachiu:
I like the idea! Here's a mock of the card layout for discussion. The sidebar's main state could be focused on **1) I'm working on** and **2) Today I learned** ... when you click on the arrow, it shifts the sidebar state to show either "# people are online learning" or "# accomplishments"

This way we can keep the home dash less crowded with information and focus on the higher priority features.

The cards would need a "Join" button as well (maybe at the top right corner or on a hover state).

The ongoing hangout cards could have "ongoing" at the top right corner.. or maybe a hover state like the "Completed" and "Cancelled" overlays.

![home-dash-card-v1 0](https://cloud.githubusercontent.com/assets/13413986/18737201/94559b9c-8043-11e6-9d4f-c0e0b1297eaa.jpg)

![home-dash-card-v1 1](https://cloud.githubusercontent.com/assets/13413986/18737204/992c0714-8043-11e6-9361-42bfe1f8e2de.jpg)

![home-dash-card-v1 2](https://cloud.githubusercontent.com/assets/13413986/18737211/a17a88fa-8043-11e6-8dcb-dae33174c490.jpg)

anbuselvan:
Awesome @adachiu.

adachiu:
@lpatmo For the number count of accomplishments, is it not a daily count? On the live site right now, it's showing the total amount of accomplishments since it started so the number of accomplishments would just keep on increasing. Did we want to consider maybe capping it at weekly or even monthly accomplishments?

[...]

PR #281, opened 2016-10-05
Issue 193 - Make hangout start/end time optional

crazcalm:
Initial commit. Added code to set the end time of created hangouts to 24 hours later (given that the
user does not choose an end time for themselves). Changed the display format so that these day long (or
greater) hangout sessions show up as an end time of "To Be Announced".

Any UI that was created or changed is a temporary placeholder. I suck at UI. Help would be greatly
appreciated


client/templates/hangout/hangout-item.js
```diff
@@ -47,17 +47,30 @@ Template.hangoutItem.helpers({
...
-      ' - ' +
-      moment(hangout.end).tz(tz).format('MMMM Do h:mm a z') +
-      ' | ' +
-      hangout.users.length +
-      ' joined';
+
+    // Formates the date based on whether or not the difference of the start
+    // and end time is equal to or greater then 24 hours (95% of 24 hours...).
+    if (startEndDiffDays < 0.95 ){
+      return moment(hangout.start).tz(tz).format('ddd MMMM Do YYYY, h:mm a z') +
```
crazcalm: Thanks for the tip! I saw your comment last night and I worked on it while I was on the train this morning. As a result, I didn't have the internet so I was unable to see the variable names you used. So, mine is slightly different. 

When you have a moment, look at my commit and tell me what you think.

camilleriluke: Welcome @crazcalm, 👍 on the changes.

I am also going to post a bit more suggestions / tips on the new code. Let me know if you do not understand something or you want to discuss something further

client/templates/hangout/hangout-item.js
```diff
@@ -47,17 +47,18 @@ Template.hangoutItem.helpers({
   },
   getDate: function(hangout) {
     var tz = TimezonePicker.detectedZone();
+    var startEndDiffDays = (hangout.end - hangout.start)/(1000*60*60*24)
+    var startTime = moment(hangout.start).tz(tz).format('ddd MMMM Do YYYY, h:mm a z')
+    // If endtime is in 24 hours (95% of 24 hours), then show "To Be Announced". Else show the end time
+    var endTime = startEndDiffDays < 0.95 ? moment(hangout.end).tz(tz).format('MMMM Do h:mm a z'): 'To Be Announced '
+    var joinedUsers = hangout.users.length + ' joined'
```
camilleriluke: Continues with the comment above:

The `joined` string should be moved as close as possible to the template where it is being used, in this case, in the template literal. 

Reason behind this: so when you take a peek at the template you would easily predict what the output would look like. Right now if you don't have any knowledge of how the code in this function looks like you could assume that the return would be something like this: `20:00 - 21:00 | 4`

distalx: `getDate` , creates a string literal similar to this`Tue October 4th 2016, 3:18 am IST - October 4th 6:18 am IST | 3 joined ` which is being used by template. While I was working on #225 I need to use `getDate` for `/hangouts` view, and I realise that we are using it at multiple places, so I had done little bit of refactoring and converted it into `registerHelper` to reduce code repetition.

client/templates/hangout/hangout-item.js
```diff
@@ -47,17 +47,18 @@ Template.hangoutItem.helpers({
   },
   getDate: function(hangout) {
     var tz = TimezonePicker.detectedZone();
+    var startEndDiffDays = (hangout.end - hangout.start)/(1000*60*60*24)
+    var startTime = moment(hangout.start).tz(tz).format('ddd MMMM Do YYYY, h:mm a z')
+    // If endtime is in 24 hours (95% of 24 hours), then show "To Be Announced". Else show the end time
+    var endTime = startEndDiffDays < 0.95 ? moment(hangout.end).tz(tz).format('MMMM Do h:mm a z'): 'To Be Announced '
```
crazcalm: The ternary does make that line long and awkwardly looking. I can change that. I will probably change it to something like:

`var endTime = value1`
`if (condition){`
`    endTime = value2`
`}`

Is that okay?

client/templates/hangout/hangout-item.js
crazcalm: I can move the " joined" into the template.

In terms of the naming convention of the getDate method, because I am not sure of all the places that it is used. I am new to meteor and javaScript and even though seems like there is a 1 to 1 ratio between the templates and the js files, I do not know if that is true. If this is true, then I have no problem changing the name of that method.

I am also not sure what to call it (given that I changed the name). If you have suggestions, feel free to pass them along.

crazcalm:
Thanks to  @distalx I was able to put the end hangout button on the hangout page. I am currently having trouble with the Meteor.call function. I am trying to do use the edit-hangout method and I have most of the values, but the start time value is giving me trouble. On the hangout page, the start time is in a string format. Is there an easy way to convert that to a Date format?

Also, for the edit-hangout function call, do you need to send all the values? If possible, I would like to send just the end-time and the something to identify the corresponding hangout in the database.

crazcalm:
@camilleriluke I was in the middle of fixing the merge conflicts when I realized that edit hangout call has a new data point that has to be passed (description_in_quill_delta). I do not think it is possible for me to obtain this value on the hangout page. Do you have any suggestions of what I should try?

client/templates/hangout/hangout-item.js
camilleriluke: 👍

camilleriluke:
@crazcalm 

> I do not think it is possible for me to obtain this value on the hangout page. Do you have any suggestions of what I should try?

You mind adding some links to the code please (you can also commit your current changes), so I can better understand and help you out.

crazcalm:
@camilleriluke Sorry for the wait. I merged master into this branch so all the code was up to date and push it to this branch.

The main issues were that the templating for the getDate was moved from the js file to the html file (hangout.js and hangout.html). Similar changes were also made to other pages that once held that code.

Also, I was originally using the editHangout meteor call when the end hangout button was clicked. The requirements for that call were changed and I am no longer able to use it. It should be noted that I should not be using that call for the end hangout button. I spoke with @distalx for a bit and he gave me some tips on how to correctly perform that call. I will try to implement that on Monday and push the changes so you can see them.

@camilleriluke The main question I have for you is how should I go about changing the html and js files so that (when an end time is not specified for a hangout event) the "To Be Announced" shows on the page in the location where the end time of the hangout event would normally go.

lpatmo:
@crazcalm If I'm not mistaken, in the new homepage cards redesign in #203, we don't have an end time showing up anywhere. So I think we can put off the question for now. :) But in theory, assuming that the "end" time for an undefined event is always set 24 hours after the "start" time, we can create a helper function named `tba` in the javascript that returns `true` if this is the case -- and then, in the HTML, use `{{# if tba }} <h3>No End Time Set</h3> {{ else }} ... {{/if}}` to control what the user sees.

By the way, I haven't looked in your branch for quite a while. I finally have it checked out now, though, and I want to help with the select box UI. Will update more tomorrow, hopefully!

lpatmo:
I pushed up a commit that changes the "set end time?" link into a select box with options:

- End manually
- 1 hour
- 1.5 hours
- 2 hours 
- 3 hours
- Custom end time

The datetimepicker for the end time is always visible right now, mostly for debugging purposes.

All looks well when I select one of the options and change the start time. However, I noticed two issues:

1. If I change the start time to a date in the future -- and then back to a date in the past, the end time does not update automatically with the correct time; I have to force an update by selecting another option on the select box.

2. The end time also does not reflect the correct AM/PM value if I change the AM/PM value two times in the start time datepicker.

Feedback for commit 79b3d05 welcome! Thanks.

distalx:
I was able to fix the date time issue, and added the functionality to end the hangout,
currant available options for `select box ` are as follow 
25 minutes 
1 hours
1.5 hours
2 hours
3 hours
End manually

in case of `End manually` the end time will be set to start time + 24 hours, 

`End hangout` option will be available to, Host and moderators during the hangout.

The branch could really use some code cleanup and refactor, I have pushed the changes/fixes.

Refactoring could be addressed in new PR/issue.

crazcalm:
@distalx Thanks for the assist! The only thing that I have noticed that seems off is that I do not see the end hangout button for the events that I have created. 
screen shot 2016-10-18 at 12 03 25 pm

Once I can see the button, I can try it out. When I was implementing the end hangout button functionality, I noticed that if I passed only the hangoutId to the Meteor call, the call would work but the information on the page would disappear. As a result, I ended up passing everything attached the the "this" object to the backend. I mention that just to say look out for it. 

Our code is almost the same (I followed the outline you gave me --thanks again for that), but yours is much more polished. I did a git diff between your code and mine to try to figure out why the end button was not showing up, and I could not figure it out. If you would like to  see what I wrote, you can find it here https://github.com/crazcalm/cb-v2-scratch/tree/issue-193-end-button-still-showing.

I am going to continue off of your changes and work on the "To Be Announced" template part that is missing.

@distalx Thanks again for all the help!

distalx:
@crazcalm `End Button` will only appear if `hangout` document has an `url` field. When We start a hangout by clicking on a `start` button `google` sends us back the `hangout url`. While running app locally - we wont be able to receive any `callback` form `google` because in our `public/app.xml` we have specified callback url as follows
`var callbackUrl = 'https://codebuddies.org/callback/hangout/'+postID;` which points to the `codebuddies.org/callback/hangout/`. To test the `End Hangout` functionality, You can add `url` field manually to the hangout document, value could be any string.

Thanks for adding `To Be Announced` part to the templates. 

I have removed the few extra lines of code and also refactored few `helper` functions into `registerHelper` for reusability and reduced the repetitive code.
Also we had missed the  `hangout clone modal` - for optional data-time feature, I have updated that too.

crazcalm:
@distalx Cool. The part about the url makes sense. It seems like this PR is about done! I am excited!

client/templates/hangout/hangout.html
```diff
@@ -80,80 +88,95 @@ <h2 class="{{# if completed this }}completed{{/if}}">
...
 
-            {{#if isInProgress this}}
+            {{#if isHangoutInProgress start end}}
 
-              {{#if isJoined}}
+              {{#if isAttending users}}
                 {{#if url}}
+
                     <a class="btn btn-cb2 btn-grey btn-success" href="{{url}}" target="_blank">{{_ "join_long"}}</a>
+
```
sergeant-q: Could add some space between the `launch hangout` and `end hangout` **button**

sergeant-q: screen shot 2016-10-25 at 11 31 59 pm

lpatmo:
Yay!

Added the bit of spacing between the buttons per @sergeant-q's suggestion.

Otherwise looks EXCELLENT! Thanks @crazcalm and @distalx!

PR #283, opened 2016-10-05
Hangout listing is now available for visitors.

distalx:
I've added a view for visitors [non-logged in users] to browse the hangouts. #225 .

User can browse hangout by visiting `/hangouts` view/path, Also user can browse hangouts by preference, `live` and `upcoming` are currently available preferences.

Default preference is `live`.


lpatmo:
Awesome, @distalx! :) This'll be wonderful for giving non-logged-in users a snapshot of upcoming/live hangouts.

I made some UI commits.

screen shot 2016-10-06 at 2 02 59 am

There are a few more small things (for example, is it possible to have a starburst on all the "live" hangouts for consistency's sake?), but I will file them as new issues after this is merged in.

PR #287, opened 2016-10-06
Referencing images directly from the repo.

camilleriluke:
Updated references inside the readme images to point directly to the images in the repo not to the deployed code.

I would suggest to follow this pattern so that we will never end up in a not-in-sync state. This happens when a PR affects some images url, and if it gets merged but not deployed the references to the new images will be broken until the next deploy.


lpatmo:
Makes sense, @camilleriluke -- grazie!

PR #292, opened 2016-10-07
Refactor css to only use variables for colors

Briix:
Let me know if there are any inconsistencies still left in the code.


camilleriluke:
Good step in the right direction 👍

lpatmo:
Thanks, @Briix!

PR #296, opened 2016-10-07
Fix hangout cards on frontpage

Briix:


lpatmo:
For some reason I'm having trouble doing a `git checkout issue-259-fix-hangout-cards-frontpage`. I see the error `error: pathspec 'issue-259-fix-hangout-cards-frontpage' did not match any file(s) known to git.` Is anyone else having the same problem?

Briix:
@lpatmo That's weird. Just tried it on a different machine and it works fine for me. 
Maybe try to either do `git fetch` first or try to checkout a new branch from the remote one, like `git checkout -b issue-259-fix-hangout-cards-frontpage origin/issue-259-fix-hangout-cards-frontpage`

abhiii5459:
@lpatmo  @Briix  +1 for git fetch. Maybe your list of branches isn't updated yet

lpatmo:
@Briix @abhiii5459 Thanks for the tips. Wasn't working still in my old clone (perhaps because I'd accidentally created a blank branch with the same name??), so I re-cloned and got it working that way. Anyway, it looks great -- merging in now; thanks again Simon.

PR #306, opened 2016-10-08
Fix Issue 300: Design Issue login confusion for new users who click to a single post page

RichardTran:
Made the fix to change Original to New to make it clearer to Sign into Slack. Change OK to 'Sign in with Slack'. After clicking the confirmation button, it mimics the sign in button in the navbar.

Branch: issue-300

Original:
image

New:
image


sergeant-q:
Good job on the changes @RichardTran.  I think we may also need to apply the same updates to the edit hangout function as well. If you click on the `Edit` button on the top right, it is currently displaying this:

screen shot 2016-10-10 at 8 55 00 pm

Can we replicate the change in this functionality as well?

lpatmo:
@RichardTran Thank you for working on this! (And glad the dev environment setup finally started working.)

@sergeant-q made a good point that this popup happens in a couple of different place. @distalx suggests that this could be a global helper function in `templates/main.js`. Does that make sense?

RichardTran:
Yes that makes sense. I'll try to get this done today, and move the function into main.js

lpatmo:
I have to admit I'm stumped on how to move the contents of 

```
confirmButtonText: TAPi18n.__("sign_in_with_slack"),
          type: 'info'		          type: 'info'
      });		 +      },
       function(){
         var options = {
           requestPermissions: ['identify', 'users:read']
         };
         Meteor.loginWithSlack(options);
       }
     );
```

into a global helper function, since this sweetAlert only happens after an event. @distalx, can you help?

distalx:
Updated PR. ⬆️

i18n/en.i18n.json
```diff
@@ -79,6 +79,9 @@
   "remove_owner_from_hangout": "Hey! You can't remove yourself from your hangout.",
   "delete_hangout_confirm": "Are you sure?",
   "delete_hangout_text": "Note: All participants will be notified via email of cancellation",
+  "continue_popup_text":"By continuing, I agree that I am at least 13 years old and have read and agree to the terms of service and privacy policy.",
```
sergeant-q: I think we should slightly reword this to:

**By continuing, I agree that I am at least 13 years of age and have read and agreed to the terms of service and privacy policy of this site.**

lpatmo:
Whoops, didn't notice that you'd worked on this, @distalx. Thanks. And thanks @sergeant-q for reviewing. Will suggest/commit some additional changes soon.

sergeant-q:
@lpatmo I spoke with @distalx and also let him know I'll be pushing some additional updates to this branch.  To avoid any conflicts, I'll push mine first then we can review. Is that ok?

lpatmo:
@sergeant-q Sounds good; thank you!

i18n/en.i18n.json
distalx: @sergeant-q Updated PR. Check it out and feel free to let me know if there is anything else I missed.

lpatmo:
Note: had to run `meteor npm install --save lodash` in my terminal afterwards to start up the app.

PR #308, opened 2016-10-09
Fixes Issue 286. Effectively displays the time left for the hangout to start on Slack notification

### LINKED ISSUE #286: Edited - Changing this to countdown slack alert times.  (Old title - Time displayed on Slack for hangout alerts is in UTC as opposed to the user's respective timezone)
Discussed this with @lpatmo . 
I think this can be remedied by obtaining the timezone of each individual user using Slack APIs and adjusting dynamically.
This is a vague description. I'll add more to it.

Self-assigning to attempt to fix it.

abhiii5459:
Fixes #286 
Like so - 

screen shot 2016-10-09 at 9 22 36 pm

I created two commits by mistake. The second overrides the first, so merging shouldn't cause any issues.


abhiii5459:
PRs 318, 319 have been created instead.

PR #315, opened 2016-10-10
Change footer position and height

dantesolis:
Hey guys, the footer *sitcks* on all pages except those with content, which will scroll with it. Don't know if that was the aim? Let me know.


dantesolis:
👍

PR #319, opened 2016-10-10
Fixes issue 286 : slack notifications with time left

abhiii5459:
Fixes issue #286 .

Slack notifications now have the time left in terms of days/hours/ minutes as applicable.

screen shot 2016-10-09 at 9 22 36 pm

### DIFF
--- a/server/slack/methods.js
+++ b/server/slack/methods.js
@@ -27,6 +27,52 @@ slackNotification = function(hangout, type){
      pretext = `A new *${hangout.type}* hangout has been scheduled by <@${hangout.host.name}>!`;
   }
 
+  let date_start, now_value, now_formatted_time, difference, minutes;
+  date_start = moment.utc(hangout.start);
+  now_value = (moment.utc()); //.format('MMMM Do YYYY, h:mm a z');
+  now_formatted_time = (moment.utc()).format('MMMM Do YYYY, h:mm a z');
+  difference = (moment.duration(date_start.diff(now_value)));
+  minutes = difference.asMinutes();
+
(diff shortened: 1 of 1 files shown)


server/slack/methods.js
```diff
@@ -27,6 +27,52 @@ slackNotification = function(hangout, type){
...
+     if((minutes >= 0 && minutes < 1))
+     {
+       var time_left = 'Hangout starts now!';
+     }
+     else {
+       var time_left = (rem+1)+' mins left';
+     }
+  }
+
+  else if(minutes<0){
```
sergeant-q: Would there ever be a situation where we either create a new hangout or send out a reminder for a hangout that already ended? Trying to understand the use case for this piece of code. 

Currently - our cron job runs **hourly** for reminder notifications.

abhiii5459: Right. I was trying to cover all bases, but I just noticed that we can't even schedule a hangout in the past. As such, this case will never be hit, but I've added it for the sake of completeness. If deemed unnecessary, I will go ahead and remove it.

sergeant-q: I don't think we need this case at this time and no need to prematurely add one. Less checks this function has to do! Let's remove it.

abhiii5459: Doing it right away.

abhiii5459: @sergeant-q  Done.

lpatmo:
Looks great. Thanks @abhiii5459 for working on this and @sergeant-q for reviewing!! Merging in!

PR #320, opened 2016-10-10
View user number has different font fix

ericathedev:
Related Issue: codebuddiesdotorg/cb-v2-scratch#309


lpatmo:
Looks great, @ericathedev! Thank you!

PR #329, opened 2016-10-13
Fix/issue 52

sharynneazhar:
Let me know how I can make this better  :) 

I couldn't figure out how to add the additional profile info to the existing profile object without overwriting it, so I just set it to a new key called `profileInfo`.


client/templates/profile/profile.js
```diff
@@ -4,6 +4,10 @@ Template.profile.onCreated(function(){
   DocHead.setTitle(title);
 });
 
+Template.registerHelper('editMode',function(){
```
distalx: Would you please replace it with the template helper.? avoid use of `registerHelper` helper if it's being used in single template.

server/users/methods.js
```diff
@@ -4,6 +4,22 @@ Meteor.methods({
...
+      location: String,
+      bio: String,
+      website: String,
+      twitter: String,
+      github: String,
+      facebook: String,
+      linkedin: String
+    };
+    check(profileInfo, pattern);
+    Meteor.users.update({_id: Meteor.userId()},
```
distalx: Before updating a record, confirm logged in user. 
and to update the profile object without overwriting, you can use dot-notation to access and set fields inside objects. `{$set: { profile.twitter: profile.twitter, profile.github:github ... }`

sharynneazhar:
updated ⬆️ 😊

client/templates/profile/profile.html
```diff
@@ -0,0 +1,83 @@
...
+                    {{#if status.online}}
+                      <a><i class="fa fa-circle online-now" aria-hidden="true"></i></a>
+                    {{/if}}
+                  {{/if}}
+                </h3>
+                <p>
+                  <i class="fa fa-map-marker" aria-hidden="true"></i>{{profile.location}}
+                </p>
+                <div class="social-icons">
+                  <a href="https://twitter.com/{{profile.social.twitter}}" target="_blank">
```
distalx: can you wrap  `optional` elements like location, twitter, etc... with `if` block ? 
so that could only displayed when those element has some value in it.
i.e. `{{#if profile.social.twitter}} <a>...</a> {{/if}}`

client/templates/profile/profile.html
```diff
@@ -0,0 +1,83 @@
...
+                    <i class="fa fa-linkedin" aria-hidden="true"></i>
+                  </a>
+                </div>
+              </div>
+            {{/with}}
+          </div>
+          <div class="col-md-6 profile-top-right">
+            <p class="bio">{{userInfo.profile.bio}}</p>
+            <a href="{{userInfo.profile.website}}" target="_blank">
+              {{userInfo.profile.website}}
```
distalx: can you move `website` with other social elements, and use either `globe` or `link` icon for it ? no need to modify it, in `schema`.

adachiu:
@sharynneazhar Requesting some layout adjustments to enhance UX hierarchy.

@distalx Yesterday we discussed about having the bio text under the name and having the social links on the right side.

- normally we read from left to right so people would be reading the bio first before the social links (the social links are secondary). 
- If we haven't set a character count yet to the bio, I suggest we keep the bio limit at 140 characters (shown in mocks below). That way, it is a quick and concise bio, and it makes the layout look more clean.

@distalx The green dot could be on another issue as a feature. It shows if the user is online or not.

v1
dash-bio-v1

v2
dash-bio-v2

I'm leaning towards v2. Thoughts?

sharynneazhar:
I like number v2 as well. 

Also, I was thinking of maybe tweaking editing social links to have a dropdown instead. So user can choose which social platform they want to add. Right now, it's hard set to Facebook, Twitter, GitHub, LinkedIn, and their own website. Thoughts? I can create a new issue/PR for this enhancement.

sharynneazhar:
The green dot should show up for other users and if it's the current user's profile, the edit profile button would show up in place of the green dot. Although, I do like how v2 is set up right now to show both. Great use of the white spawe. 😊

distalx:
@adachiu after we discussed, I tried to put together a mockup for that, and I come up with something similar to your version 2,  I had couple of concern, with that, 
i.e. left and right side might look horizontally uneven in following case.
no bio, no social info and only location.
no bio, no location and only social info.

so I thought we could have a new issue to discus that.

@sharynneazhar dropdown for social info is defiantly sounds appealing - but I believe it will add more complexity on the backend part. if you want to give it a try let's open a issue for that.

And to display user status ( green dot ) we just have to display `user.status` property on the template. it seems like a minor change to me. it can be address here or feel free to create new issue for that.

adachiu:
@sharynneazhar How does the editing social links currently look like? I like the idea of letting the users add more social platforms! That would be great to create a new issue for it.

RE green dot: Thanks for clarifying! I think it works either way; having the green dot beside edit button vs not having it beside.

@distalx we could align the social links to the right:
screen shot 2016-10-16 at 12 13 01 am

screen shot 2016-10-16 at 12 13 49 am

or

v3 (not the best)
dash-bio-v3

v4
dash-bio-v4

v5 (Might make the page too long, but if we keep the bio character count at 140 limit, then it should be alright)
dash-bio-v5

adachiu:
@distalx @sharynneazhar We could actually remove the location. It doesn't seem too important to know a user's location, unless we have in-person hangouts.

distalx:
@adachiu if we remove the location then almost all the mocks looks good.

lpatmo:
Just getting to reviewing now... this is awesome, @sharynneazhar! :) 

Noticing a bug, though, with the links: if I only have my `github` link filled out and nothing else, nothing shows up when I save. And if I only have `facebook` filled out, Facebook, LinkedIn, and Github will icons will all show up even though I didn't fill out my Github and LinkedIn info.

Aha -- it's because `{{# if profile.social.facebook}}` was put in there a couple of extra times. I committed this change:

```
+++ b/client/templates/profile/profile.html
@@ -37,12 +37,12 @@
                       <i class="fa fa-facebook" aria-hidden="true"></i>
                     </a>
                   {{/if}}
-                  {{# if profile.social.facebook}}
+                  {{# if profile.social.github}}
                     <a href="https://github.com/{{profile.social.github}}" target="_blank">
                       <i class="fa fa-github" aria-hidden="true"></i>
                     </a>
                   {{/if}}
-                  {{# if profile.social.facebook}}
+                  {{# if profile.social.linkedin}}
                     <a href="https://www.linkedin.com/in/{{profile.social.linkedin}}" target="_blank">
                       <i class="fa fa-linkedin" aria-hidden="true"></i>
                     </a>
```

Also made the following changes:
- fiddled with the CSS for the "Edit Profile" button a little
- Added a default Slack icon linking to http://codebuddiesmeet.slack.com/messages/{{username}}

Please `git pull`! Thanks. :)

sharynneazhar:
I will work on some of the tweaks to match the UI/UX @adachiu have. 

I will also create a separate issue and PR for social links enhancement.

lpatmo:
@sharynneazhar Sounds good; will wait to merge this PR until the design tweaks. Excited!

sharynneazhar:
Updated PR. Check it out and feel free to let me know if there is anything else I missed.

lpatmo:
Excellent, @sharynneazhar! Merging in. Thank you!

PR #330, opened 2016-10-13
fixe for Issue 324

### LINKED ISSUE #324: [Deployment question] 'Cannot find module' error on cbv2-staging.herokuapp.com (our test staging site)
The error below didn't show up in `localhost:3000`, but showed up in http://cbv2-staging.herokuapp.com. If this were on localhost, I normally would do a `meteor npm install` to fix the problem. But not sure how to fix this on a live heroku instance.

screen shot 2016-10-11 at 1 11 14 am

distalx:
The workaround is to add `quill.js` locally.

close #324


lpatmo:
Thanks for this! Sorry it took me so long to merge.

PR #342, opened 2016-10-17
commented on error catching function

cflatts:
Commented on an error throwing function for Hacktoberfest.

### DIFF
--- a/server/email/methods.js
+++ b/server/email/methods.js
@@ -25,6 +25,7 @@ emailNotification = function(hangout, type){
     subject: subject,
   }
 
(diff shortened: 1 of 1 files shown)


sergeant-q:
@abhiii5459 this is technically true that this is not a function but a block statement.  We've discussed the validity of this PR since it's simply adding a comment to our code base. We do encourage better code comments but this may not qualify since the comment is inaccurate. 

@cflatts do you want to update your comments to reflect what this piece of code is and it's purpose?

sergeant-q:
I am closing this PR due to no follow up and it has been agreed upon that the code is self explanatory.

PR #354, opened 2016-10-18
Fix/issue 353 footer placement

sharynneazhar:


lpatmo:
Looks great -- thanks, Sharynne! Only one issue I see (we can file a separate issue for this too): on the profile page, the footer blocks the "load more" button on the first column.

sharynneazhar:
@lpatmo I'll fix that. I didn't notice that bug because I have 0 hangouts in the column 😅

lpatmo:
@sharynneazhar If case it's helpful, you can seed your localhost app with some fake hangouts by changing `  "seeder":false,` to `  "seeder":true,` in your settings-development.json and restarting the app.

Issue #360, opened 2016-10-23
Privacy page UI


adachiu:
Updating the privacy page column layout so it isn't squished to one side.

PR #361, opened 2016-10-23
Issue 360 privacy ui

### LINKED ISSUE #186: Create a terms of service page
URL: http://codebuddies.org/terms-of-service

Include this:
```
General notice: If you're under 13, it is illegal for you to have an account on Slack because of the Children's Online Privacy Protection Act (https://en.wikipedia.org/wiki/Children%27s_Online_Privacy_Protection_Act). Instead, please check out https://code.org/learn, or https://www.commonsensemedia.org/lists/coding-apps-and-websites. Alternatively, you can ask a parent to create an account and post on your behalf. But without express parental permission, we're not allowed to collect information from you — i.e. you can't have an account here, or we'll be in legal trouble. Sorry!
```

adachiu:
Update column layout and spacing between paragraphs to match other pages like FAQ and About.

Closes issue #360 

I see that Children Online Privacy Protection Act has already been included so I think we could also close #186

### DIFF
--- a/client/css/_about.scss
+++ b/client/css/_about.scss
@@ -20,7 +20,7 @@
 
   .about-middle {
     background: $bg-white;
-    padding: 100px 0;
+    padding: 50px 0;
     h2 { @include font-face(700, 32px, bold, $primary-grey); }
     .flags { margin: 20px 0;}
     img { margin: 10px;}
--- a/client/css/_about.scss
+++ b/client/css/_about.scss
@@ -50,6 +50,10 @@
     }
   }
 
+  h2 {
+    margin: 50px 0 30px;
+  }
+
   h3.popover-title { margin-top: 0px;}
   h3 { font-weight: 200;}
 }
--- a/client/css/_about.scss
+++ b/client/css/_about.scss
@@ -50,10 +50,11 @@
     }
   }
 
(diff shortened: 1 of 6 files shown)


adachiu:
@lpatmo thanks for letting me know that! Should the global margin-bottom rule be placed in `_base.scss` ?

lpatmo:
`_base.scss` works! The thing is, we might want to check that having the margin-bottom on all h3s doesn't make the `about`, `faq`, etc. pages look weird. If it does break something, could create a _privacy.scss file specifically for the HTML beneath `div#privacy` (after changing id="about" to id="privacy" on line 2 in privacy.html). 

Anyway... don't mean to make this more complicated than it has to; just noting! Let me know if any of what I said didn't make sense. Thanks.

adachiu:
@lpatmo That makes sense. Let me update this before you merge! 😄  I'll update the `faq` and `about` spacing too with SCSS and delete the `<br>`

lpatmo:
Thank you! :)

adachiu:
@lpatmo Updated! Removed most `<br>` after adding the a global for both `h2` and `h3`. I also updated the `faq`, and `about` pages since the global affected those pages as well.

PR #379, opened 2016-10-28
Fixes on top/bottom area padding & .home-psst mobile

### LINKED ISSUE #374: Extra padding on landing page & mobile padding/width
1) When accessing the site on a Macbook Pro in Chrome, Firefox, and Safari, there's two sections on the landing page that have extra padding (top and bottom, respectively) that may not be needed. Here's the two sections, as screenshot recently on codebuddies.org:

screen shot 2016-10-27 at 2 20 32 am

----------

screen shot 2016-10-27 at 2 22 00 am

To point it out specifically, I'm talking about the extra blue background on top of "Learn and chat about code!" in the first screenshot, and the extra blue background underneath "codebuddiesmeet.slack.com team" in the second one. Using Chrome DevTools to take out the padding, the two then look like this:

screen shot 2016-10-27 at 2 30 25 am

----------

screen shot 2016-10-27 at 2 30 05 am

With two changes to padding in each respective area, the content at the top of the landing page isn't pushed down anymore, and the background above and below the second screenshot's content is evened out, cleaning up the design some. Also checked on my iPhone Chrome app, same issue in it. 

2) While checking the site out on my phone, I noticed the following:

screen shot 2016-10-27 at 3 01 00 am

There's no left/right padding for the paragraph that starts with "On the CodeBuddies site," and the "Join Hangouts" example image is too small to view and read properly. I also went into DevTools and believe I found where these can be fixed.

I'd like to work on these two things if it looks all right to do so. Let me know! 🙂

kindlingscript:
Fixes #374.

Taking @adachiu's advice, I cut the top padding in half from 120px to 60px so it now looks like this:

screen shot 2016-10-27 at 10 08 17 pm

Padding at the very bottom blue area was removed, and this is how the mobile padding on the .home-psst section is looking on Chrome DevTools:

screen shot 2016-10-27 at 10 20 08 pm

screen shot 2016-10-27 at 10 18 47 pm

screen shot 2016-10-27 at 10 17 48 pm

If the media queries need refactoring to fit a style or something I missed, let me know. I was going by resizing my browser width and adding those two in when the content broke (image again grew too small to really read, needed a bit more left/right padding). Did just notice that I copied the 20px on h3, p padding twice where the second instance should be 40px, so I'll fix that up now and push the commit up (the images above reflect the 40px on phones, not the original 20px). 😅

### DIFF
--- a/client/css/_home.scss
+++ b/client/css/_home.scss
@@ -1,7 +1,6 @@
 #home {
   background: $cta-blue;
   padding-top: 50px;
-  padding-bottom: 120px;
 
   p { line-height: 27px; color: $primary-grey; }
 
@@ -56,7 +55,7 @@
   }
 
   .home-top, .home-ask {
-    padding: 120px 0 0 0;
+    padding: 60px 0 0 0;
     background: $dodgerblue;
     overflow: hidden;
     h2 {
@@ -233,5 +232,18 @@
     }
   }
 
+  @media only screen and (max-width: 695px) {
+    .home-psst {
+      h3, p { padding: 0 20px; }
+      .hangout-example { width: 65%; }
+    }
+  }
+
+  @media only screen and (max-width: 530px) {
+    .home-psst {
+      h3, p { padding: 0 20px; }
+      .hangout-example { width: 80%; }
+    }
+  }
 
 }
--- a/client/css/_home.scss
+++ b/client/css/_home.scss
@@ -200,6 +200,10 @@
     p { line-height: 27px; margin-top: 40px;}
     #intro {
       h3 {@include font-face(600, 24px, bold, $primary-grey);}
+      h3, p {
+        @include breakpoint(small) { padding: 0 20px; }
+        @include breakpoint(xsmall) { padding: 0 40px; }
+      }
       text-align: center;
       max-width: 550px;
       margin: 0 auto;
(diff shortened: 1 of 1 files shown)


kindlingscript:
A few minutes ago it occurred to me that you all probably have a breakpoints file, so I went and found it. I haven't dealt with this part of SCSS / Sass before, but I Googled and found the following article: https://responsivedesign.is/develop/getting-started-with-sass

A little bit down the article gives this sample code:

```
@include breakpoint(breakpointname) {
		/* Styles */
	}
```

Plus:

```
.grid-1-4 {
	width: 100%;
		@include breakpoint(phablet) {
		width: 50%;
	}
	@include breakpoint(laptop) {
		width: 25%;
	}
}
```

I'm still learning this, but to me it looks as if with this, `@media` shouldn't be necessary within the specific modules, but `@include...` instead. So I'll get to refactoring this with the project's breakpoints. 👍

lpatmo:
@kindlingscript Yes, exactly! :) You probably already discovered this in _breakpoints.scss... we have four breakpoints that we can include inside our CSS selectors:

 @include breakpoint(tiny) { ... }
 @include breakpoint(xsmall) { ... }
 @include breakpoint(small) { ... }
 @include breakpoint(medium) { ... }
 @include breakpoint(large) { ... }

PR #389, opened 2016-10-29
Update footer.html

BennyCarlsson:
Fixes #<387>.

Added a $ icon to the footer that links to opencollective.com/codebuddies


lpatmo:
Thank you for this PR, @BennyCarlsson! Closing because we updated the spec on issue 387, but we appreciate the PR.

PR #394, opened 2016-10-29
Ellipsis Character Number Display on Slack and Homepage

foxiae:
No issue number. Discussed in #codebuddies-meta.

Ellipsis would show in both the Slack notifications no matter what for descriptions (at 201 character limit) and for title (at 101 character limit _only in Slack_). 

The proposed fix alleviates confusion if the character text is supposed to continue on. 

Note: I wasn't able to get the italics working on the hangout description for Slack. I'm not sure why, but it does provide a nice visual difference of which hangouts have more text than the ones under 201.

Also - not sure how we're versioning our files here. It looks like @sergeant-q had some one filled out for the Slack notification. Let me know if I should bump it up to 0.0.2.

### DIFF
--- a/client/templates/hangout/hangout-item.js
+++ b/client/templates/hangout/hangout-item.js
@@ -31,7 +31,12 @@ Template.registerHelper("hangoutOwner", function(ownerid){
 
 Template.hangoutItem.helpers({
   getDescriptionTruncated: function(description) {
-      return description.substring(0,201)+"...";
+      if(description.length > 201){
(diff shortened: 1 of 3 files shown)


foxiae:
Let me clean this up, actually. I created this branch off of another branch I'm work off of, so that commit got stuck in this one.

foxiae:
Closing in favor https://github.com/codebuddiesdotorg/cb-v2-scratch/pull/396 for a cleaner PR.

PR #408, opened 2016-10-31
Template.instance(...).loadHangouts runtime error fix

camilleriluke:
Fixes this error:
`Exception in template helper: TypeError: Template.instance(...).loadHangouts is not a function`

Was getting it occasionally when refreshing the page.


lpatmo:
Thanks for this, @camilleriluke! I wasn't able to recreate the error on the home page, sadly. Do you mind clarifiying where you saw it?

sergeant-q:
I can confirm I was also seeing this error in the console and the hangouts did not load. This is only happening in Chrome browser for me.

lpatmo:
So sorry for the tremendously late PR acceptance here, @camilleriluke!

PR #413, opened 2016-10-31
Fix issue 404: “mouseover glitch”

### LINKED ISSUE #404: Odd hover behavior on contributors section of About page
hover

I think the mouseover event trigger should have 20-30px buffer around the pop-up, so you don't have to follow the pseudo element to access links, and the behavior is a little more predictable.

Possible fix could be using an "invisible box" as a parent div and tying the event actions to that.

Tested on Chrome 54.

kevincoleman:
Fixes #404.

I juggled the margin & padding a bit to make sure there was no cold zone between the image and the popover dialog.

Also fixed a typo in `package.json`

### DIFF
--- a/client/css/_about.scss
+++ b/client/css/_about.scss
@@ -41,7 +41,13 @@
     display: inline-block;
   }
   a.user-popover {
-    .popover { margin-top: -10px;}
+    padding-top: 10px;
+    .img-circle {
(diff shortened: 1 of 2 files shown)


lpatmo:
Thank you @kevincoleman!

Issue #321, opened 2016-10-11
Volunteers descriptions are just lorem ipsum


gytdau:
On the bottom of https://codebuddies.org/about - under the heading `And all those who've made a positive impact on the project and the community`, everyone's popup box is a paragraph of filler text and a generic username. The links don't go anywhere meaningful either.

Did we originally plan to have a system where people fill this info in?

lpatmo:
@gytdau Heh, yes -- sorry, that's on me! I have to finish contacting everyone and make sure they're okay with their descriptions / notes about how they've helped.

gytdau:
Oh, okay. That sounds difficult to coordinate - maybe we should have a unified form for everyone?

adachiu:
@lpatmo I created a Google Form for collecting CodeBuddies volunteer descriptions. Shared it with you already. Perhaps you can pin the link to `codebuddies-meta` channel using the `@channel` so you don't have to ask each volunteer to submit information.

lpatmo:
Thanks @adachiu! Maybe we can add it to `contributing.md` as one of the last steps, for those who want their information added to the `Contributors` list.

foxiae:
@lpatmo I think that goes along with what we all talked about in Slack a couple of weeks ago. I'll template something up in contributing.md and submit it in.

foxiae:
Actually... I was thinking. If we have people edit both files, when they go do their PR, wouldn't there be a merge conflict? (Just seems like it'll produce a huge headache.)

Would a possible workflow be:
1. Create branch
2. Work and commit stuff
3. If it's been less than 3 days (just an example), merge master changes into your branch
4. Submit PR
5. Once PR has been approved, merge master changes again
6. Edit contributing files
7. Merge branch into master

lpatmo:
Makes sense generally, though I have a question about step 6 -- are you suggesting that contributors add their information to the /about page themselves, instead of submitting their info to the google form Ada started?

foxiae:
I was also thinking we could just have a standing branch for contributing files, and they could just commit there. Then we can do periodic releases of it.
Yes. We can template out the HTML and have them update the information - with instruction. (This is for new contributors. The ones existing on the page now, the Google Forms will work and we can paste the information in.) It would save work on us inputting the stuff in and all we have to do is review the work. If it gets *too* overloaded, we can do something like how my local Rails group handles the community: http://bmoreonrails.org/. The user profile images are randomly generated on refresh.

PR #414, opened 2016-10-31
Issue 321- Adding Yourself as Contributor/Other Refactoring

foxiae:
Addresses #321 .

@lpatmo This is my proposed instructions for new contributors adding themselves as a contributor (see the section "How do I add myself as a contributor?").  I think it's most stream-lined process to handle it.  

I've also added a TOC and have done a bit more formatting to make the file more readable.


lpatmo:
Wow, this is comprehensive! Thank you so much, @foxiae!

My only question/concern is whether or not it's actually possible to create multiple PRs using the same `adding-contributor` branch. I guess we'll have to test it out, or if it turns out the answer is no (even after fresh commits have been submitted) we could always ask contributors to name the branch `adding-contributor-[name]`.

foxiae:
I was thinking about that. Another option was thinking is that they can branch off adding-contributors. I was just trying to avoid as many merge conflicts as possible. If it doesn't work, we'll change it.

Issue #300, opened 2016-10-08
Design Issue: login confusion for new users who click to a single post page


lpatmo:
This is a starter issue, highly inspired by the great job of the folks at [Hoodie](https://github.com/hoodiehq). 👍
#### 🎃🙀👕🔭😻 Hacktoberfest: Free T-Shirt from DigitalOcean for Your Open-Source Contributions!

If you haven’t yet, [sign up for Hacktoberfest](https://hacktoberfest.digitalocean.com/) to earn an exclusive T-Shirt, which you can win after making four open pull requests.
#### 🤔 What you will need to know to fix this issue

HTML/CSS. You'll need to edit the popup. Please see: `templates/hangout/hangout.js` line 67.
#### 🎯 The Goal

This popup is confusing for new users who click to a hangout post and try to RSVP to the hangout:

![screen shot 2016-10-08 at 9 29 41 am](https://cloud.githubusercontent.com/assets/4512699/19214471/ed23232c-8d39-11e6-91de-52f01ec83521.jpg)

We need to make it clearer that participants need to click on the "Sign in with Slack" button the top right of the page to log in.

Also make it clear that they need to be logged into the codebuddiesmeet.slack.com Slack team already in order to log in, and if they don't have an account, they can sign up for one on the codebuddies.org home page.

The proposed solution: 
On the popup, put instead:
You are almost there!
"Sign in with Slack" button
"Get your Slack invite" if you don't have an account with codebuddiesmeet.slack.com
## 📋 Step by Step
1. 🙋 Claim this issue: Comment below. Please 🙏 only claim if you want to start working on it within a day. 
2. Please read our [contributing.md](https://github.com/codebuddiesdotorg/cb-v2-scratch/blob/master/contributing.md) document to get started. You'll need to run the app locally, and create a branch for this issue.
3. Fix the issue/add the feature
4. Open a PR to master branch, mentioning this issue number. Simply opening a valid PR counts for your Hacktoberfest participation!
5. Someone will review your code, and once it's approved we'll merge it and deploy it to codebuddies.org.

RichardTran:
I can work on this issue

ryzokuken:
He claimed it seven days ago... If he hasn't been working on it, could I?

lpatmo:
@ryzokuken Sorry, should've updated this thread... there is an open pull request for this issue (https://github.com/codebuddiesdotorg/cb-v2-scratch/pull/306) that is in progress.

[...]

PR #419, opened 2016-11-14
Fixed additional lodash related issues

sergeant-q:
Fixes lodash compatibility issues and issue #300 

Due to dependency of lodash for QuillJS, we were not able to use the latest version of `lodash` npm package. Reverted the changes back. 

Additional updates are specific to issue #300 and now all actions/buttons will prompt any visitor to log into Slack first before proceeding.


lpatmo:
Thanks @sergeant-q!

PR #420, opened 2016-11-15
Remove word-break from status

newswim:
Hi CodeBuddies!

So, I was posting a status update to the "Today I Learned..." section, and noticed the word-break was making the paragraph look a bit difficult to read. This PR will turn this:

screen shot 2016-11-15 at 1 13 05 pm

INTO THIS:

screen shot 2016-11-15 at 1 35 51 pm

Hope this is an improvement :)

### DIFF
--- a/client/css/_status_list.scss
+++ b/client/css/_status_list.scss
@@ -52,10 +52,10 @@
     }
 
     .status {
-      padding-left: 10px;
+      padding: .4em;
+      background: white;
       display: inline-block;
       @include font-face(400, 15px, normal, $primary-grey);
-      word-break: break-all;
     }
 
     .status-date {


lpatmo:
Thanks for the PR, @newswim! One question: does this fix accommodate for what happens if someone pastes in a URL in their learning message?

newswim:
Hmm not sure! I'll look into that tomorrow, happy to keep this open until we know. It would be nice to add syntax highlighting as well, but that may be a bit more involved.

newswim:
@lpatmo is the current issue that a link is not generated?

lpatmo:
I was wondering if pasting in a long link would cause the text to overflow to the right without `word-break`, but have just tested it out, and it looks okay! So... never mind. :)

(Not an issue about generating links)

PR #445, opened 2017-01-30
Basic addition of Jitsi confrencing code.

BethanyG:
Works On #\<364\>.

Added code to Hangout details page (html and js) to hook up the Jitsi API for hangouts.  Also changed browser_policy.js to allow jit.si (but not jitsi.org or jitsi.com yet).


sergeant-q:
Good stuff @BethanyG.  I noticed as soon as you enter the hangout details page, Jitsi video fires up automatically with video on. I'm sure this is not a desired behavior. Do we have options to disable video or provide the ability to 'launch' jitsi when one desires?

lpatmo:
Worked on this with @BethanyG. A couple of things on the roadmap for this PR:

1) pass in {{_id}} of the post into the `room` string:
```
 <script>
                    var domain = "meet.jit.si";
                    var room = "starflower123";
                    var width = 700;
                    var height = 700;
                    var api = new JitsiMeetExternalAPI(domain, room, width, height);
                </script>
```
2) Only have this embedded hangout load on the page IF 
     a) the user has RSVPed to the hangout;
     b) when the hangout is in progress

lpatmo:
The bullet points above have been implemented, if anyone would like to check out the branch `issue-364` and review.

lpatmo:
According to https://github.com/jitsi/jitsi-meet/blob/master/doc/api.md Jitsi lets me overwrite a config.js setting to add `startVideoMuted: 0` (meaning every participate after the 0th will join with video muted); have pushed that up.

lpatmo:
Some of the other big features are kind of dependent on this `jitsi` issue (364) if we don't want messy merge conflicts.

The proposal: we launch Jitsi, but continue to have an option to launch a google hangout nearby in case Jitsi is still a pain after their recent bug fixes.

PR #459, opened 2017-02-22
Sort upcoming hangouts from soonest to furthest - fixes #458

### LINKED ISSUE #458: Move hangouts that are in progress but not yet completed to the top
What is the current behavior?
Hangouts on the homepage are listed, presumably in order of creation, but with the furthest date at the top.

What is the expected/_wished for_ behavior?
Hangouts should be sorted by completion date/time, with the closest future date at the top, and all hangouts-to-be above completed ones

All

newswim:
This should resolve the order issue. I just created another helper which checks if the template is completed. This seemed like the most non-envasive. Would appreciate more eyes :microscope: -- thanks!!

### DIFF
--- a/client/templates/hangout/hangout-list.js
+++ b/client/templates/hangout/hangout-list.js
@@ -16,12 +16,12 @@ Template.hangoutList.onRendered(function() {
 
     instance.loadHangouts = function() {
       var now = new Date();
-      return Hangouts.find({'start': { $gte : now }}, {sort: { start: 1 }});
+      return Hangouts.find({'end': { $gte : now }}, {sort: { start: 1 }});
     }
 
     instance.loadCompletedHangouts = function() {
       var now = new Date();
-      return Hangouts.find({'start': {$lt : now}}, {sort: { start: 1 }});
+      return Hangouts.find({'end': {$lt : now}}, {sort: { start: -1 }});
     }
 
     instance.addMoreHangouts = function(){
--- a/client/templates/hangout/hangout-list.js
+++ b/client/templates/hangout/hangout-list.js
@@ -21,8 +21,7 @@ Template.hangoutList.onRendered(function() {
 
     instance.loadCompletedHangouts = function() {
       var now = new Date();
-      var now_in_min = now.setMinutes(now.getMinutes())
-      return Hangouts.find({'end': {$lt : now_in_min}}, {sort: { start: -1 }});
+      return Hangouts.find({'end': {$lt : now}}, {sort: { start: -1 }});
     }
 
     instance.addMoreHangouts = function(){
--- a/client/templates/hangout/hangout-list.js
+++ b/client/templates/hangout/hangout-list.js
@@ -15,8 +15,13 @@ Template.hangoutList.onRendered(function() {
     var instance = this;
 
     instance.loadHangouts = function() {
+      var now = new Date();
+      return Hangouts.find({'start': { $gte : now }}, {sort: { start: 1 }});
+    }
 
-      return Hangouts.find({}, {sort: {start: -1}});
+    instance.loadCompletedHangouts = function() {
+      var now = new Date();
+      return Hangouts.find({'start': {$lt : now}}, {sort: { start: 1 }});
     }
 
     instance.addMoreHangouts = function(){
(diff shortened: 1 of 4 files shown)


stain88:
Gonna be picky here, but maybe the completed hangouts should be in reverse order, so most recently finished is at the top

client/templates/hangout/hangout-list.js
```diff
@@ -15,8 +15,13 @@ Template.hangoutList.onRendered(function() {
...
 
     instance.loadHangouts = function() {
+      var now = new Date();
+      return Hangouts.find({'start': { $gte : now }}, {sort: { start: 1 }});
+    }
 
-      return Hangouts.find({}, {sort: {start: -1}});
+    instance.loadCompletedHangouts = function() {
+      var now = new Date();
+      return Hangouts.find({'start': {$lt : now}}, {sort: { start: 1 }});
```
stain88: I would say these could be sorted in reverse

client/templates/hangout/hangout-list.js
```diff
@@ -21,7 +21,8 @@ Template.hangoutList.onRendered(function() {
 
     instance.loadCompletedHangouts = function() {
       var now = new Date();
-      return Hangouts.find({'end': {$lt : now}}, {sort: { start: -1 }});
+      var now_in_min = now.setMinutes(now.getMinutes())
+      return Hangouts.find({'end': {$lt : now_in_min}}, {sort: { start: -1 }});
```
newswim: Good catch @lpatmo ! Thank _very much_ for chasing that behavior down.

newswim: I think think we need to have checks for both somehow, since only hangouts which have been manually ended will have their `end` time in minutes. This might generally be the case, but any seeded hangouts will now not show up.

Can we change the format of hangouts that are ended manually to have a regular timestamp? Then I think the previous code would work.

newswim:
@stain88 @lpatmo

I tried updating the server-side method to maintain the same date format, let me know if this works 🔨 🔨 🔨 🔨

newswim:
All seems good on my end, newly ended hangouts get placed at the top of the Completed pile.

stain88:
Looks good indeed

lpatmo:
Yay teamwork! Thanks for fixing, @newswim. :) One thing I noticed: `const end = date.setMinutes(date.getMinutes() - 1);` could probably be `const end = date.setMinutes(date.getMinutes());`, since we're subtracting the minute in the next line? But not a big deal.

newswim:
Good point!

newswim:
hey @lpatmo - what's your reasoning for squashing all of these commits into one? Just curious.

stain88:
Keeps things cleaner in a large code base - 1 commit per issue is nicer than 10 small ones

newswim:
:(

It would be nice to have a note about that in the contrib.

lpatmo:
> Keeps things cleaner in a large code base - 1 commit per issue is nicer than 10 small ones
Yep, exactly! (Or in some cases, 100 small ones.) "Squash and merge" is the default button when I click to merge in a PR:
screen shot 2017-02-24 at 8 39 56 am

@newswim We can add a note about it in contrib for sure. Are you unhappy about squash for some reasons?

newswim:
I realize that my feelings on this are probably a result of the GitHub reward model, but I think when people contribute (whatever the amount), it's nice at least to see that reflected in our commit history.

I understand the desire to keep the commit history clean, but I think the consequent of having a slightly more muddled project commit history is worth it for both A.) The granularity of being able to see every commit, and B.) The reward (green contribution square) to a person's record.

I'm curious of other people's views on this.

Also I don't think the logic is entirely consistent if you just go back a few pages in the commit history.

stain88:
In reply to point A, they're all still available in the `reflog`. For point B, when contributing to open-source, I would say quality is more important than quantity. There are several articles arguing both sides, but most I've seen tend to fall on the side of squashing.

lpatmo:
@newswim Huh, I didn't realize that the PR merge type affected your commit history too!

I've read more articles that err on the side of squashing too, but I'm not unopposed to foregoing squashing in the future if it's a small number of commits.

stain88:
Just to elaborate a bit on point B, what I meant to say was, if someone is crawling your github account to see what you've contributed to open source, they would probably care more about whether you've fixed an issue, than how many goes it took you to fix it at first glance.

newswim:
Either way is reasonable, I just personally would like to know ahead of time. Also, I would argue that it _does_ play in to the experience of contributing, especially for first-timers.

@stain88 that's pretty debatable. The point is to have a scannable history, which includes things like _when_ people are most active, _how_ they got from the problem to the solution.. that's all lost when you squash.

PR #489, opened 2017-03-25
Allow anonymous users to see the logged-in home page (new "Hangouts" link on nav)

### LINKED ISSUE #467: Separate out the "home logged in" from the "home logged out" templates
This will have the side effect of fixing the reload problem, where users see a glimpse of the logged-out home page view before they see the logged-in view. 

But I'm starting to think we should make the logged-in home page view available for non-users, so that they can browse the hangouts that have occurred in the past.

This is NOT a priority issue, but thought I'd raise this for discussion.

lpatmo:
Fixes #467 

- created a new template `hangouts-logged-in`
- This new template points to /hangouts, which is also a new link on the nav
- Clicking on the CB logo will bring you back to the "original" home page that non-logged-in users saw.
- Anonymous users can now see the list of upcoming and completed hangouts
- Anonymous users *cannot* see who else is online until they log in
- Anonymous users can see past learnings
- If you're not logged in but sign in via the popup, you will be redirected to /hangouts
- The popups are now standardized if you try to do something that only logged-in users are able to do (like go to a profile, upvote a learning, or schedule a hangout). 

This is a pretty major change, so review/feedback is very appreciated. You can check out this branch by running `git checkout issue-467` in your terminal. Please let me know if I've missed anything.


lpatmo:
Merging this to make the redo of the hangout cards easier.

PR #493, opened 2017-04-10
display url links in learning items

### LINKED ISSUE #492: Automatically turns text starting with "http" in "learnings" section into URLs
In the learnings sections of hangouts (e.g. https://codebuddies.org/hangout/c4a2WpcYX9AyTzamB), any text that starts with "http" or "https" should automatically be converted into an `<a href="X">X</a>` link.

Let's use regex and javascript to solve this instead of applying QuillJS to learnings' input boxes.

rudkovskyi:
Fixes #492

Turns text into URLs in Learning section

### DIFF
--- a/client/templates/status/learning-item.html
+++ b/client/templates/status/learning-item.html
@@ -3,9 +3,9 @@
        <li>
           <div class="kudo-holder display-inline">
               <label class="kudo-count {{#unless currentUser}} continue-popup {{/unless}}">
-              
+
                 <i class="fa fa-angle-up fa-2x btn-{{#if buttonStatus}}un{{/if}}like  btn-learning btn-xs"></i>
-             
+
               <p class="align-center">{{kudos}}</p>
             </label>
           </div>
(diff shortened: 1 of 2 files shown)


lpatmo:
Thank you for this PR, @rudkovskyi!! :) 

I think your original idea to use `Template.onRendered` i.e. 

```
 Template.learningItem.onRendered(function(){
   var pattern = /((http|https)\:\/\/[a-zA-Z0-9\-\.]+\.[a-zA-Z]{2,3}(\/\S*)?)/g
   this.$('p').html(this.data.title.replace(pattern, "<a href=\"$1\">$1</a>"));
 });
```
is necessary, because the current solution (creating a helper function) causes the links to show up like this:

screen shot 2017-04-10 at 11 39 19 pm

In other words, I'm not sure we can avoid having to change the links only *after* the text renders on the page. What do you think?

Thanks,
Linda

rudkovskyi:
That is why you have to use `{{{ }}}` instead of `{{ }}`. I am not expert here, although as far as I see `onRendered` executed only once when the page is gets loaded so it's fine until you modify the text.

lpatmo:
@rudkovskyi My apologies, I totally missed that you triple-{{{ed to escape the text!

rudkovskyi:
@lpatmo no worries! It's fine

lpatmo:
This looks perfect, thanks for this PR! :)

rudkovskyi:
Thank you! ^_^

PR #519, opened 2017-06-01
Upgraded the dependancies to the latest one

kanji-karanja:
The dependancies were abit outdated so added the new dependancies that were causing the app to crash now and then failing to start.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,18 +8,18 @@
     "test:full-app": "meteor test --full-app --driver-package=practicalmeteor:mocha --settings=settings-development.json --port 3100"
   },
   "dependencies": {
-    "babel-runtime": "6.23.0",
-    "bcrypt": "1.0.2",
-    "body-parser": "^1.15.2",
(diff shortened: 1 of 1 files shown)


lpatmo:
Thanks @kanji-karanja!

distalx:
Hi @kanji-karanja, I'm not able to reproduce this issue.
what is the exact error message you're seeing?
Thanks.

kanji-karanja:
There was an error that kept saying "App is crashing" and in the console it
would tell me to try to install something after which updating the
dependencies that were abit outdated resolved the issue and the app started
well

lpatmo:
@kanji-karanja Sorry for the super slow reply. I haven't been able to reproduce this issue either. If you could, could you pull in the latest from `staging` and run the app again? And let us know if you still see the same problem? I'm hesitant to upgrade the dependencies in case it invisibly breaks features, since we don't have any tests.

PR #528, opened 2017-06-28
Issue 517

### LINKED ISSUE #517: Update Button Issue Screen Width: ~1000-1140px
**Bug**
What is the current behavior?
**Screwed up**
**Nice and neat**
**All most likely although I am using Chrome**

akosasante:
Fixes #517 .

The `.status-actions` div had a hard set height (something like 60px) so it wouldn't adjust it's size in the narrower viewport range before the breakpoint (as per issue opener this was occuring on Screen Width: ~1000-1140px). This was leading to the "Update" button getting pushed down and kind of blending with the background, not looking very nice. Set height to auto and do some flexbox niceness to let the button push down a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ privkey.pem
 settings.json
 npm-debug.log
 my-settings.json
+.vscode/**
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@ privkey.pem
 settings.json
 npm-debug.log
 my-settings.json
-.vscode/**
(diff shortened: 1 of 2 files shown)


akosasante:
@lpatmo : Ah, I don't think I actually meant to include that; but yes that is the settings folder for Visual Studio Code editor

lpatmo:
@akosasante Got it, thanks. Going to merge this in now!

lpatmo:
Shouldn't see this problem on https://staging.codebuddies.org/hangouts, in a few minutes.

PR #530, opened 2017-07-04
User n role updates

distalx:
These changes are required to accommodate grouped roles for user management.

migration script will make the following change
from
```
"roles": [
    'user'
]
```
to
```
"roles": {
   "CB": [
       'user'
   ]
}
```

and it contains the required modification to the user-management system.


lpatmo:
Ah, so the idea is that there'll be one of three user roles?

```
"roles": {
   "CB": [
       'user'
   ]
}```
```
"roles": {
   "CB": [
       'moderator'
   ]
}
```
```
"roles": {
   "CB": [
       'admin'
   ]
}```

distalx:
@lpatmo migration won't have any effect on user's current role - we are just grouping it inside the group called 'CB'.
'CB' would be the default user group for the entire system, with this change we would be able to add/assign more groups and their roles to user as require.

lpatmo:
(Ah I see, so we can have something like:

```
"roles": {
   "CB": [
       'moderator'
   ],
  "JS Study Group": [ 'user' ], 
  "Python Study Group": ['admin']
}
```)

PR #540, opened 2017-07-21
Change URLs to the correct one

### LINKED ISSUE #539: Update Github links everywhere
Links on the site (and in our contributing.md) still point to https://github.com/codebuddiesdotorg/cb-v2-scratch; we should make it point to https://github.com/codebuddiesdotorg/codebuddies. 

Right now there's an automatic redirect so there's no breakage, but we should make the link consistent with the new github URL.

This is a `good-for-beginners` issue. :)

jlee124:
Fixes #539 .

I'm new to open source, so this is my first contribution. 

I went through all of the files and replaced https://github.com/codebuddiesdotorg/cb-v2-scratch with https://github.com/codebuddiesdotorg/codebuddies. I also went through the README.md and replaced /cb-v2-scratch with /codebuddies as well. If this wasn't needed, please let me know and I can go back and change those.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,4 +4,4 @@ Describe the big picture of your changes here to communicate to the maintainers 
 
 Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.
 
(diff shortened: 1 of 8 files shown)


lpatmo:
Hmm... actually, the issue I noticed before looks like it was a glitch. This is fine; am going to merge it in. Thanks @jlee124!

PR #550, opened 2017-08-21
I added some much-needed commas to the json file so the alerts would …

### LINKED ISSUE #549: End Hangout Button Grammar
When you end a hangout, an alert box pops up. The option to end the hangout reads "Yes end it!"; there should be a comma between "Yes" and "end it".

arturolei:
Fixes #549:

I added some much-needed commas to the SweetAlert.


sergeant-q:
Thanks for the great attention to detail!

lpatmo:
Thanks for the PR @arturolei and thanks for reviewing, @sergeant-q!

PR #604, opened 2017-09-20
Issue-603. Fix issue with margin around the label box inside the stud…

### LINKED ISSUE #603: STUDY GROUPS: Margin of owner/member box.
This should be easy for beginners. It's just adding few pixels of margin to the member/owner box at the top.

Current
screen shot 2017-09-17 at 20 34 12
screen shot 2017-09-17 at 20 34 30

Proposed - 5 pixels margin top
screen shot 2017-09-17 at 20 33 40
screen shot 2017-09-17 at 20 34 05

dantesolis:
Fixes #603 **<ISSUE-603\>**

Adds a value of `margin-top: 10px;` inside the `.label-box` class to handle margin of the member box.

**Question**: Hey guys I added some local variables to handle the `margin` on the `_study_group.scss` file. I was thinking that perhaps we could either:

1. create a `_variables.scss` file where we define and use those reusable `css` variables inside global variables, or

2. add them on top of each file, pertaining each file.

What do you guys think?

### DIFF
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -1,3 +1,6 @@
+$margin-zero: 0 !default;
+$margin-base: 10px !default;
+
 #study-groups {
   h2 {
     margin: 10px 0px;
@@ -63,6 +66,10 @@
 
   .well--end-of-page {
     padding: 5px;
+  }
+
+  .label-box {
+    margin-top: $margin-base;
   }
 }
 
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -1,6 +1,3 @@
-$margin-zero: 0 !default;
-$margin-base: 10px !default;
-
 #study-groups {
   h2 {
     margin: 10px 0px;
@@ -8,7 +5,7 @@ $margin-base: 10px !default;
   }
   h3 {
     text-transform: uppercase;
-    margin: 20px 0px 10px 0px;
+    margin: 20px 0 10px 0;
     @include font-face(700, 14px, bold, $black);
   } 
   .tab-pane h3 {
--- a/client/css/style.scss
+++ b/client/css/style.scss
@@ -1,3 +1,4 @@
+@import 'variables';
 @import 'colors';
 @import 'base';
 @import 'breakpoints';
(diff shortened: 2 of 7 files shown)


lpatmo:
Thanks for the PR, @dantesolis! 

I would vote for the first: use a `_variables.scss` file. It should be imported relatively early, though; probably not after _colors.scss? 

And btw, we have a `.margin-top-1` class in `_base.scss`; at some point we may want to remove it/fold it in with `$margin-base`.

client/css/_study_groups.scss
```diff
@@ -1,3 +1,6 @@
+$margin-zero: 0 !default;
+$margin-base: 10px !default;
```
lpatmo: Let's move these two lines into _variables.scss?

client/css/style.scss
```diff
@@ -24,4 +24,5 @@
 @import 'update_status';
 @import 'hangout_card';
 @import 'hangout_board';
-@import 'study_groups';
\ No newline at end of file
+@import 'study_groups';
+@import 'variables'; // global vars to be added to this
```
lpatmo: Let's move the `@import 'variables';` up to the very top -- before `@import 'colors';`

dantesolis:
Done 💯.  Let me know what you guys think. I could take care of the 'migration' of the position values to the `_varialbes.scss` file and then create another PR.

lpatmo:
Thanks for the new commits, @dantesolis! If you have time, could you create a new issue about `_variables.scss`? We use `.margin-top-1` and `.margin-top-2` as a CSS class sprinkled throughout the HTML fairly often, and I'm not sure we should remove the ability to use it as a class... but we can discuss in the issue thread. Maybe I'm misunderstanding what you mean by the usage of position values.

dantesolis:
@lpatmo will do. We could leave it as a class in the mean time until we know for sure how to proceed and just extend that class either with à `@mixin` or with an `@extend`. What do you think? 


Also, and this is just a thought, should we try to use some `css-vars()`- as in for variable déclarations - when possible and use `scss` for `mixin` déclarations and other nifty things.

Just an idea I had, what do you think :)

PR #607, opened 2017-09-24
feat - search for study groups.

### LINKED ISSUE #559: STUDY GROUPS: Add search feature to /study-groups
People should have the ability to search for groups on /study-groups.

See: http://staging.codebuddies.org/study-groups

distalx:
Fixes #559 

Search feat for SG and several other improvements.


client/templates/resources/resources_list.html
```diff
@@ -4,7 +4,7 @@
       {{#each resources}}
         <div class="well">
           <h4>
-            {{#if isInRole 'owner, admin, moderator' _id}}
+            {{#if isInRole 'owner, admin, moderator' ../_id}}
```
lpatmo: Is this an example of the spread operator?

distalx: we are accessing the parent scope  with the help of `../`, it means to look for `_id` into parent scope. it's a `blaze` thing, not related to es6 spread operator.
`each resources` would have it's own id field, but we are interested in studygroup's `_id`, 
with the help of that we can see that if the user is belong to one of the role mention above.

lpatmo: Thanks for the explanation! :)

PR #629, opened 2017-10-01
Issue 584

### LINKED ISSUE #584: STUDY GROUPS: Messy study groups dashboard
screen shot 2017-09-10 at 21 30 28

Elements should have the same height, and new elements should go to the left.

schoettkr:
Fixes #584 .

This fixes the messy study groups dashboard by giving them elements a fixed height and reduce bottom padding (to avoid the vertical scrollbar where it is not needed and just occurs because of the bottom padding) at a viewport that is >992px.

Also this line `@media only screen and (min-width: $medium + 1) and (min-width: $large) {` (line 29)  in _breakpoints.scss should read `... and (max-width: $large)..` so I already changed this.

Furthermore as it is the bootstrap medium breakpoint is at >=992px but the mixins start to apply at >992px (992px + 1 = 993px) so a viewport width of 992px will still have a messy dashboard ...

Feedback is appreciated

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -4,7 +4,7 @@ accounts-password@1.3.6
 acemtp:accounts-slack@1.0.9
 alanning:roles@1.2.16
 aldeed:simple-schema@1.3.3
-allow-deny@1.0.5
+allow-deny@1.0.9
 anti:i18n@0.4.3
 arillo:flow-router-helpers@0.5.2
 autoupdate@1.3.12
--- a/client/css/_breakpoints.scss
+++ b/client/css/_breakpoints.scss
@@ -26,7 +26,7 @@ $large: 1200px;
       @content;
     }
   } @else if $media == medium {
-    @media only screen and (min-width: $medium + 1) and (min-width: $large) {
+    @media only screen and (min-width: $medium + 1) and (max-width: $large) {
       @content;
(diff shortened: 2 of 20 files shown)


lpatmo:
screen shot 2017-10-06 at 12 55 41 am

```
@media only screen and (min-width: 1201px)
_study_groups.scss:1
#study-groups .well {
    height: 130px;
    padding: 19px 19px 10px;
}
```

Perhaps to fix, we could apply a `.card` CSS class to each `<div class="well">` for the study group cards on /study-groups, and then have the rules revolve around that? What do you think? 

e.g. in `client/templates/study_groups/all_study_groups.html`:
```
+ <div class="well card">
- <div class="well">
```
```
@media only screen and (min-width: 1201px)
_study_groups.scss:1
#study-groups .card {
    height: 130px;
    padding: 19px 19px 10px;
}
```

schoettkr:
Hey, thanks for your response! The study group page itself indeed does look messy I didn't notice that (and that it would have messed up other `well`s too).
I added the `card` class and attached the css rule.
All Study Groups:
all_studygroups
Sinlge Study Group:
studygroup_sidebar

lpatmo:
Thanks so much for fixing, @schoettkr! :)

PR #638, opened 2017-10-04
Feature/issue 609

dantesolis:
Fixes #\<issue-609\>

- [x] Fix issue-609 pertaining text sizes on `hangouts.html`. 
    - [ ] Create `_typography.scss` and set global values for `heading`s and `text`s tags inside of it.  
    - [ ] `html` font-size increased to `1rem === 16px`. Create variables for font-size for most common used tags.
- [ ] Tackling `a11y` issue on the next PR.

**Mobile**

mobile_hangout_cards

**Desktop***

mobile_hangout_cards

desktop_hangout_cards


lpatmo:
screen shot 2017-10-06 at 12 40 56 am
screen shot 2017-10-06 at 12 41 05 am

We should set the font-size for `.hangout-card .hangout-topic` to 1rem, maybe.

And define h4, h5 in _typography.scss to be 1rem too.

Also noticed that the "This will only take a moment" text that flashes after doing a search got quite large, too. That one is a `p.alert`.

client/css/_base.scss
```diff
@@ -4,15 +4,21 @@ html {
...
   -moz-box-sizing: inherit;
   box-sizing: inherit;
 }
 
 body {
-  padding: 0;
-  #profile { background: $bg-white;}
+  padding: $padding-zero;
+  margin: $margin-zero;
+  font: normal (#{$font-md} / 1.3) $font-stack;
```
lpatmo: Question: What do you think about continuing to use the mixin style of `@include font-face(600, $font-lg, normal, $black);`? Maybe we'll need to add a $font-stack parameter to the font-face mixin.

lpatmo:
(Added a couple of notes/questions. Thanks again for tackling this, and sorry for the slow review!)

client/css/_base.scss
dantesolis: Good idea, I was afraid of using that mixin without asking for "permission" first, but I'll add a new param to it.

lpatmo: @dantesolis Thanks! Let me know if you need any help. :)

nalbina:
What is needed for this PR to complete?

dantesolis:
@all, currently working on this. :)

PR #639, opened 2017-10-04
Check study group name is unique before new study group can create

### LINKED ISSUE #585: STUDY GROUPS: Don't allow to create a study group with the same name
Self explanatory. I was able to create two study groups with the same name.

railsstudent:
Fixes #585 .

I query study_groups collection in MongoDB to perform exact match on title and visibility equals true and return record count. If count is greater than 0, then Meteor throws error. Otherwise, the function follows the same logic to create new study group.

New study group is allowed to create if the name is used for the first time or all study groups with the same name are archived.

### DIFF
--- a/server/study_groups/methods.js
+++ b/server/study_groups/methods.js
@@ -20,6 +20,12 @@ Meteor.methods({
        throw new Meteor.Error('StudyGroups.methods.createNewStudyGroup.not-logged-in', 'Must be logged in to create new Study Group.');
      }
 
+     // Check study group name is unique
+     isTitleAlreadyUsed = StudyGroups.find({title: data.title, visibility: true}).count() > 0;
+     if (isTitleAlreadyUsed) {
+       throw new Meteor.Error('StudyGroups.methods.createNewStudyGroup.title-already-used', 'Study group name is already used');
+     }
+
      const user = Meteor.user();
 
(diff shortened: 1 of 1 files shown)


railsstudent:
selection_008

server/study_groups/methods.js
```diff
@@ -20,6 +20,12 @@ Meteor.methods({
        throw new Meteor.Error('StudyGroups.methods.createNewStudyGroup.not-logged-in', 'Must be logged in to create new Study Group.');
      }
 
+     // Check study group name is unique
+     isTitleAlreadyUsed = StudyGroups.find({title: data.title, visibility: true}).count() > 0;
```
distalx: - Please, use block scope variable, `const  isTitleAlreadyUsed`
- use `findOne` instead of `find`, `find` will scan the entire collection- where `findOne` will exit once it find the first matching record.
- mongodb queries are case sensitive so we should use regex as a workaround here, 
`{ title: /^data.title$/i, visibility: true }`

distalx:
@railsstudent, Thanks for the PR! Here are some suggestions I hope you'll find helpful.

railsstudent:
@distalx  Fixed. Please kindly review

lpatmo:
LGTM too!

PR #657, opened 2017-10-12
Issue 597 study group status

### LINKED ISSUE #597: Give each group its own "status" area
When I open up a study group page, it would be nice to see:
- who else is online
- a place to type in my current static status within the study group (e.g. "just starting out with this tutorial", or "on Ch. 5" or "interested in pairing")

This should go in the sidebar.

railsstudent:
Fixes #597 .

This PR makes two changes in single study group.
In the sidebar,  Online Members section is created to display members that are currently online.
In Members tab, members can input their current status and display it in the member table below. Furthermore, latest status retains in textarea to save typing in case the text demands little change
members_nonownermode
members_ownermode


server/members/method.js
```diff
@@ -0,0 +1,17 @@
+Meteor.methods({
+  editStatus: function(data) {
+    check(data, {
+      user_id: String,
+      status: String,
+      study_group_id: String
+    });
+
+    if (!this.userId) {
```
distalx: Here, We should also check if user belongs to the SG or not.
```
const actor = Meteor.user()
if (!actor || !Roles.userIsInRole(actor, ['owner','admin', 'moderator', 'member'], data.studyGroupId )) {
   ...
```

server/members/method.js
```diff
@@ -0,0 +1,17 @@
...
+      throw new Meteor.Error('Members.methods.editStatus.not-logged-in', 'Must be logged in to edit the member status.');
+    }
+    StudyGroups.update({_id: data.study_group_id, 'members.id': data.user_id },
```
distalx: while modifying info related to current user, we should either use `Meteor.user()` or `Meteor.userId()` for security reasons. `'members.id': Meteor.userId()  },`

server/members/method.js
```diff
@@ -0,0 +1,17 @@
+Meteor.methods({
```
distalx: This method could be moved to either SG/methods.js or users/methods.js

client/templates/study_groups/study_group_online_members.js
```diff
@@ -0,0 +1,14 @@
+Template.studyGroupOnlineMembers.helpers({
+  onlineUsers: function(){
+    const onlineUsernames =  Meteor.users.find({ "status.online": true }, { fields: { username: 1 }})
+                        .fetch()
+                        .map(m => m.username);
```
distalx: we should use _id, it will be globally-unique, `m._id`

railsstudent: done

server/members/method.js
railsstudent: move to SG/methods.js

server/members/method.js
railsstudent: fixed

server/members/method.js
railsstudent: Throw 403 access error

railsstudent:
@distalx  Please kindly review

distalx:
Looks good to me, thanks a lot @railsstudent .

PR #672, opened 2017-10-19
added danger red color to text/border when no description

### LINKED ISSUE #670: CSS: Highlight the "Edit Description" error with a red border and red text if group description has not been filled out yet
When someone creates a study group for the first time, their eyes might skip over the need to click on "Edit Description" to fill out a goal and a description for their group.

Simple UI enhancement: let's make the text red, and put a red border around the area so that anyone who has permission to edit the description area will automatically be drawn to the area.

screen shot 2017-10-18 at 8 25 59 pm

If you're a new contributor to this project, please check out docs.codebuddies.org.

The red color is $danger-red, which is defined in _colors.scss.

kennyhuynh125:
Fixes #670 .

Added the danger red to the color of the text and border of the div where the No description text is showing up.

### DIFF
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -34,6 +34,8 @@
   	margin-right: 5px;
   }
   .well, .jumbotron {
+    color: $danger-red;
+    border: 1px $danger-red;
   	margin: 0px 0px 10px 0px;
   }
   .card {
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -36,10 +36,8 @@
   .well, .jumbotron {
   	margin: 0px 0px 10px 0px;
   }
-  #description-box {
+  .warning-well {
     border: 1px solid $danger-red;
-  }
-  .warning {
     color: $danger-red;
   }
   .card {
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -34,9 +34,11 @@
   	margin-right: 5px;
   }
   .well, .jumbotron {
-    color: $danger-red;
     border: 1px $danger-red;
   	margin: 0px 0px 10px 0px;
+  }
+  .well p {
+    color: $danger-red;
   }
   .card {
     @include breakpoint(medium) { height: 130px; padding: 19px 19px 10px; }
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -34,8 +34,10 @@
   	margin-right: 5px;
   }
   .well, .jumbotron {
-    border: 1px solid $danger-red;
   	margin: 0px 0px 10px 0px;
+  }
+  #description-box {
+    border: 1px solid $danger-red;
   }
   .warning {
     color: $danger-red;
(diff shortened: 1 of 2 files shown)


lpatmo:
Feedback:
- Instead of applying the color rule to .well p, let's give the div.well that contains the "Your study group doesn't have a description yet" its own CSS class -- maybe call it `.warning` -- and apply the styles there in _study_groups.scss?

This way we can ensure that other text won't be affected, like:
screen shot 2017-10-18 at 10 37 38 pm

The other thing I noticed is that even though apply a border color, it's not showing up. I believe you're missing a `solid` argument; it needs to be `1px solid $danger-red`.

Hope this feedback helps!

kennyhuynh125:
Hi @lpatmo,

Thank you for the feedback! I just made some changes, hopefully it works now! Please take a look! :)

lpatmo:
@kennyhuynh125 That's a good step... but are you sure you wanted to make all .well .jumbotron border red? :)

screen shot 2017-10-18 at 10 58 01 pm

One possibility is putting the `.warning` class on the div itself, and then doing something like:
```
  .warning {
   border: 1px solid $danger-red;
    p { 
      color: $danger-red;
     }
   }
```

Note: I have not tested the above; might be worth previewing it in the console or running the app. Let me know if you have questions. 

P.S. Let me also know if you're having trouble getting the app to run, or if any of the instructions on docs.codebuddies.org are confusing. Installing meteor shouldn't take too long, but the first time you hit `meteor --settings settings-development.json` it takes a few minutes.

kennyhuynh125:
Hi @lpatmo 

Thank you for the feedback! I wasn't sure whether well was a class used only in that template, seems like it is used elsewhere as well. I added an ID to that div so that the red border is specific only to that div.

kennyhuynh125:
Oops! i actually didnt see the second half of your feedback :(, i believe the id will still work, but your method sounds better. i will try and work on this more tomorrow!

lpatmo:
@kennyhuynh125 Yeah, an id will technically work, but the reason I suggested naming the class `warning` and applying it to the div is so that in future, we'd be able to reuse the class whenever we had a similar well we wanted to draw attention to. :) Thanks for working on this!

kennyhuynh125:
Hey @lpatmo 

I am having trouble with getting the app to run, been getting an error. I updated meteor with a patch to see if it works again.

I tried doing the CSS on Codepen and realized that your way will include a border within the p element as well (assuming that p also has a class called warning). I was wondering if it would work if I give both the div and the p element .warning but use the other classes to style it, something like...
`.well .warning {
  border: 1px solid $danger-red;
}
.text-center .warning {
  color: $danger-red;
}`

lpatmo:
Sorry for the slow followup, @kennyhuynh125! Pushed up some changes (https://github.com/codebuddiesdotorg/codebuddies/pull/672/files#diff-00f4fbcbe51fb5f21b5fa99939627f80) that simplifies things; in short, if you apply a class to the wrapping div, you can style the border and also the text inside it too. :) Let me know if you have any questions. Thanks again for your time on this!

Issue #683, opened 2017-10-21
Easy fix: Rename`Resources` to `Links`


lpatmo:
The list of resources that study group admins can compile is really a list of links; let's call it by it's name, perhaps?

Task: rename "Resources" to "Links"

![screen shot 2017-10-21 at 2 06 55 am](https://user-images.githubusercontent.com/4512699/31850313-b7d3f350-b604-11e7-9340-9b4199b6e85c.jpg)

Relevant file: 
i18n/en.i18n.json
```141:  "resources": "Resources",```

Note: let's not worry about changing the name on the backend, but we may want to update the h2 here as well:
```
client/templates/study_groups/edit_study_group_resources_modal.html
1:<template name="editStudyGroupResourcesModal">
2-  <div class="modal fade">
3-    <div class="modal-dialog modal-lg">
4-        <div class="modal-content">
5:          <form  class="updateStudyGroupResources">
6-
7-            <div class="modal-header">
8-              <button type="button" class="close" data-dismiss="modal" aria-hidden="true">×</button>
9:              <h2>Resources</h2>
```

===
New contributors: please check out [docs.codebuddies.org](http://docs.codebuddies.org). And if you'd like to claim this issue, please be sure to leave a comment here first. Thanks!

kristinakarnitskaya:
I'll take this :)

lpatmo:
@kristinakarnitskaya It's yours! Let us know if you have questions.

PR #685, opened 2017-10-21
Easy fix: Rename to  #683

kristinakarnitskaya:
Fixes #\<issue number\>.

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue above.

Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

Always follow the [contribution guidelines](https://github.com/codebuddiesdotorg/codebuddies/blob/master/contributing.md) when submitting a pull request.

### DIFF
--- a/client/templates/study_groups/single_study_group.html
+++ b/client/templates/study_groups/single_study_group.html
@@ -80,7 +80,7 @@
             <li role="presentation"><a href="#hangouts" aria-controls="hangouts" role="tab" data-toggle="tab">{{_ 'hangouts'}}</a></li>
             <li role="presentation"><a href="#members" aria-controls="members" role="tab" data-toggle="tab">{{_ 'members'}} ({{ members.length }})</a></li>
             <li role="presentation"><a href="#learnings" aria-controls="learnings" role="tab" data-toggle="tab">{{_ 'learnings'}} ({{ learningsCount }})</a></li>
-            <li role="presentation"><a href="#resources" aria-controls="resources" role="tab" data-toggle="tab">{{_ 'resources'}} ({{ resourcesCount }})</a></li>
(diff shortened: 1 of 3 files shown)


i18n/en.i18n.json
```diff
@@ -138,7 +138,7 @@
   "members": "Members",
   "learnings": "Progress Log",
   "links": "Links",
-  "resources": "Resources",
+  "links": "Links",
```
stain88: Does links need to be in here twice?

kristinakarnitskaya: You're right. Will fix it

client/templates/study_groups/single_study_group.html
```diff
@@ -80,7 +80,7 @@
             <li role="presentation"><a href="#hangouts" aria-controls="hangouts" role="tab" data-toggle="tab">{{_ 'hangouts'}}</a></li>
             <li role="presentation"><a href="#members" aria-controls="members" role="tab" data-toggle="tab">{{_ 'members'}} ({{ members.length }})</a></li>
             <li role="presentation"><a href="#learnings" aria-controls="learnings" role="tab" data-toggle="tab">{{_ 'learnings'}} ({{ learningsCount }})</a></li>
-            <li role="presentation"><a href="#resources" aria-controls="resources" role="tab" data-toggle="tab">{{_ 'resources'}} ({{ resourcesCount }})</a></li>
+            <li role="presentation"><a href="#resources" aria-controls="resources" role="tab" data-toggle="tab">{{_ 'links'}} ({{ resourcesCount }})</a></li>
```
kristinakarnitskaya: Shall I change all the attributes/links from 'resources' to 'links' or just the name of the link?

lpatmo:
Thanks for the PR @kristinakarnitskaya and thanks for reviewing, @stain88!

PR #686, opened 2017-10-21
Show form error messages inside modal

### LINKED ISSUE #568: Show form error messages inside modal, not outside of it
Good feedback from @osanseviero:

image

For good user experience, try to contain things were they belong. Here the website is giving feedback on the form outside the modal, which seems a bit messy. This could potentially cause a11y issues as well.

agatac:
Fixes #568.

I couldn't use the current alert implementation with `Bert.alert()`. Bert attaches the alert on a different level than the modal resides. I couldn't come up with any solution to easily 'move it' inside the modal. Instead, I created a simple alert object in the template and used Bootstrap (which is already being used in the project) to style it. Let me know what you think of this implementation.

EDIT: added a screenshot.
localhost-3000-study-group-test-group-rv9ngbfowlmmj7mbj

### DIFF
--- a/client/templates/study_groups/edit_study_group_resources_modal.html
+++ b/client/templates/study_groups/edit_study_group_resources_modal.html
@@ -29,6 +29,12 @@
               {{else}}
                 <button type="button" class="btn btn-default btn-block disabled"><i class="fa fa-refresh fa-spin"></i> {{_ "state_processing"}}</button>
               {{/unless}}
+
+              {{#if alert}}
+                <div class="margin-top-1 align-center alert {{alert.type}}">
+                  <span>{{alert.message}}</span>
+                </div>
+              {{/if}}
             </div>
 
           </form>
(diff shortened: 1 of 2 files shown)


angelocordon:
That looks great, though I think UX-wise, the alert should be at the top?

agatac:
I think this can be confusing if the user is already pointing their mouse at the form field - the form jumps down when the alert appears. Especially annoying when you're about to click on the input field and it moves up/down ;) If you think this won't be an issue I will move the alert to the top.

lpatmo:
@agatac Thanks again for taking a look, and again, apologies for not being very descriptive when I originally filed the issue! For form validation, I was thinking more of something like this:

screen shot 2017-10-22 at 12 19 30 pm

If you want to pass on this issue, no worries -- let me know!

agatac:
Alright, I made some changes
capture

nalbina:
What's the status here?

distalx:
@agatac, Sorry for letting this PR linger for so long. 
Thank you for the review, @angelocordon, @nalbina , @lpatmo

PR #688, opened 2017-10-23
feat (progress log) - users can keep track of their learnings and share with other members of SG.

### LINKED ISSUE #642: Remove 140-character limit on learnings and support code markdown (or WYSIWG)
Within each study group on http://staging.codebuddies.org/study-groups, we have a textbox that lets users share what they learned or accomplished. 

screen shot 2017-10-06 at 1 13 27 am

Right now the message is limited to 140 characters, but 140 characters is too limiting if I'm pasting in code snippets.

Ideally, we'd support the same single or triple backticks commonly used on Slack (```) and (`). 

We've used QuillJS for our WYSIWG elsewhere -- for example, on the create-a-hangout modal. Not sure if we should use QuillJS here as well, or use Meteor's native `markdown` package: https://docs.meteor.com/packages/markdown.html. What do folks prefer?

distalx:
Fixes #642

### DIFF
--- a/server/progress_logs/publications.js
+++ b/server/progress_logs/publications.js
@@ -24,6 +24,6 @@ Meteor.publish("progressLogsByStudyGroupId", function(limit, id){
 Meteor.publish("progressLogsById", function(id){
   check(id, String);
 
-  return ProgressLogs.findOne({_id:id, visibility:{$ne:false} });
+  return ProgressLogs.find({_id:id, visibility:{$ne:false} });
 
 });
--- a/server/progress_logs/publications.js
+++ b/server/progress_logs/publications.js
@@ -1,0 +1,29 @@
+Meteor.publish("progressLogsByStudyGroupId", function(limit, id){
+  check(limit, Number);
+  check(id, String);
+
+  let query = new Object();
+  query['study_group.id'] = id;
+  query.visibility = {$ne:false};
+
+  let projection = new Object();
+
+  projection.fields = {"title" : 1, author: 1, study_group: 1, created_at: 1 };
+
+  let options = new Object();
+  options.reactive=false;
+  options.limit = limit;
+  options.sort = {
+    created_at: -1
+  }
+
+  return ProgressLogs.find(query, projection, options);
+
+});
+
+Meteor.publish("progressLogsById", function(id){
+  check(id, String);
+
+  return ProgressLogs.findOne({_id:id, visibility:{$ne:false} });
+
+});
--- a/client/css/_base.scss
+++ b/client/css/_base.scss
@@ -123,6 +123,10 @@ a:hover {
   margin-top: 20px;
 }
 
+.padding-0 {
+  padding: 0px;
+}
+
 .well {
   margin: 10px 0px;
   overflow: auto;
--- a/client/css/_quill_editor.scss
+++ b/client/css/_quill_editor.scss
@@ -1,3 +1,7 @@
 .ql-container[data-editor-host] {
   height: 150px;
 }
+
+.ql-container[ data-editor-progress-log] {
+  height: 250px;
+}
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -137,3 +137,19 @@
     }
   }
 }
+
+/*Progress Log*/
+.log {
+  padding: 20px 0px;
+  h2 {
+      @include font-face(700, 34px, normal, $black);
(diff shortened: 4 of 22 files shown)


lpatmo:
@distalx Sorry for the slowness looking at this. I pushed up some styling improvements (more standardized and easier-to-read on-the-page alerts and info boxes, and cursor pointer for edit/delete). 

I'm still struggling with naming: does `Progress Log` sound too serious, and would prevent folks from sharing something they learned or accomplished? If the goal were to get more people to treat this as a note-taking or learning tracker, should we rename it back to something like `Learnings` or `Notes`? Screenshots of each of the potential options below:

screen shot 2017-10-26 at 10 22 17 pm
screen shot 2017-10-26 at 10 22 53 pm
screen shot 2017-10-26 at 10 23 12 pm

lpatmo:
Also enabled tweeting of progress logs; see `b1b6f3d496dc2b5c90479b04c31b5aefe4973da5`

Initially the `_id` of the newly created progress log isn't visible, so I grabbed the latest one matching the title/slug/description in the DB using `{sort:{$natural:-1}}`.

```
    var newProgressLog = ProgressLogs.findOne({'title': data.title, 'description': data.description, 'slug': data.slug, 'visibility':{$ne:false}}, {sort:{$natural:-1}})
    tweetLog(newProgressLog);
```

lpatmo:
... and depending on what terminology we decide, the tweet will either say:

`New note logged in the [ ] group: ...`
`New learning logged in the [ ] group: ...`
`New log in the [ ] group: ...`

distalx:
Thanks for enabling auto-tweets for logs. Should we add a sort of notice on the log submission modal that this will be tweeted by CB's twitter account?

**re: `_id `**

`ProgressLogs.insert(progressLog);` will return an `_id` of newly created record.  

Very roughly:

~const progressLog = {~
```
let progressLog = {
[...]
}
const logId = ProgressLogs.insert(progressLog);

progressLog._id = logId

tweetLog(progressLog);
```
I've update that.

**RE : terminology**

I thought that Initially, we are creating this functionality for the organiser or host of the hangout to log the short summery of the hangout, so if some member miss the hangout they can easily catch up with the group by going through these log entries. for that _use case_ `progress log` or `logs` both are fine. But if I have to pick from the `learnings` and `notes` I'd pick `notes` because learnings are already being used for #TodayILearned feat.

server/progress_logs/publications.js
```diff
@@ -0,0 +1,29 @@
+Meteor.publish("progressLogsByStudyGroupId", function(limit, id){
+  check(limit, Number);
+  check(id, String);
+
+  let query = new Object();
+  query['study_group.id'] = id;
```
sergeant-q: Any reason why we're using bracket notation here and dot-notation everywhere else? We should be consistent.

sergeant-q:
@lpatmo I think **notes** feels more appropriate to me in this context. That's my vote.

server/progress_logs/publications.js
nalbina: Looking forward to your PR fixing all other bracket notations

distalx: the schema of the **progress log** is as follows.
```
{
  _id : String,
  title : String,
  study_group : [Object]
}

```
So if we want to fetch **progress log** by a study groups our find query would be against the _field_ `study_group.id`.

`query['study_group.id'] = '22'` will result in `{study_group.id: "22"}`

here JS runs toString() on square brackets so we could use special characters into our keys.
with `dot-notation` that does not seem to be feasible.

julianjohannesen:
Hi guys,

A couple of things:

1. If Learnings are going to be auto-Tweeted, shouldn't there be some sort of disclaimer to that effect? I didn't know that that was a thing until after I entered my learning notes. I'm easy going and I really don't mind, but I can see how this could really piss someone off. Should I log that as an issue?

2. Personally, I think "Notes" is a bit more inviting than "Progress Log."

3. It would be really great if the text editor recognized html and markdown. Of course, that's easy for me to say, because I have no idea how difficult it might be to implement something like that. :)

lpatmo:
@julianjohannesen On #1: It'd be great if you could log an issue on it! Thank you!

On #2: Noted. :) 

On #3: 100%. I think we do support wysiwyg here (similar to what you see when you start a hangout).

sebbel:
@distalx sup?

PR #702, opened 2017-11-04
My account alias login pass key sync

zavedctg:
Without individual sign in to get I back from the first day of my Yahoo and Nokia account email automatically.

Fixes #\<issue number\>.

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue above.

Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

Always follow the [contribution guidelines](https://github.com/codebuddiesdotorg/codebuddies/blob/master/contributing.md) when submitting a pull request.


sergeant-q:
Unless I'm missing something, this seems like an invalid PR.

PR #714, opened 2017-11-15
Feature/issue 594

### LINKED ISSUE #594: Design Request: Contrast of Hangout Tips text
The current text in hangouts tips is a bit hard to read. It's also an a11y issue.

screen shot 2017-09-12 at 23 00 19

The two ways to go are:
- Increase the contrast
- Increase the size

Or both.

The "silent hangout  started by" text is a bit hard to read as well.

dantesolis:
Fixes #594.

**Description**: Increase the font-size and the font-weight on hangout status and hangout tips


dantesolis:
No problem @lpatmo.

We could do it like that, what do you guys prefer? I thought about `BEM` conventions and how to make the `css` more object oriented. 

Like that we could, we could still target only those particular `li`s having particular class, in case there are other `li`.  But like I said what do you guys prefer :) ?

angelocordon:
I also think our SCSS needs some huge refactors (included setting up some guidelines that we can all agree on 😅)

I find [rscss](http://rscss.io/) easier to follow but I think that should be up for discussion with everybody else. Similar to BEM, just a little bit more intentional and without the varying `-` and `_`

PR #717, opened 2017-11-17
Feat - Availability tab

### LINKED ISSUE #599: STUDY GROUPS: 'Scheduler' tab
@distalx created a demo of a scheduler that could be helpful for study groups, that'll let users see what days (and times) other people are free, and the overlap between users.

Feel free to play around with the demo here:
https://tan-wombat.herokuapp.com/jam/QPt5tYoTJAawznEEo

screen shot 2017-09-13 at 9 52 49 pm

A user would:
- select a day and a length of time in their own timezone (e.g. 8-9pm EST) and click "set"
- That length of time would appear as a chunk on the graph, and overlap with other users' times. Everyone's times will be automatically converted on the backend. 
- TODO: Display on the graph the particular day that we're looking at, maybe as tabs. 

Feedback welcome! Would you find this tool useful to schedule times that other people in a study group are available? What do you think about the design that could be improved?

distalx:
Fixes #599 

- an attempt to simplify the hangout scheduling process between group members.


lpatmo:
This is awesome, @distalx. 

One HTML thought (optional): maybe instead of using div col mds for the rows/display of each time slot, maybe it could be a striped table? (See: what we did in the `members` tab)

I would be very curious if others had minor thoughts/suggestions for this feature. Here are some screenshots:

screen shot 2017-11-18 at 4 08 25 am
screen shot 2017-11-18 at 4 08 34 am

One thing that we don't clarify on the UI right now is that the time will automatically change depending on the user's computer clock. :) So there's no need for anyone to make calculations about what my time at 9am is for another person in another country. 

I'm not sure how exactly we could clarify that time conversions are automatically calculated regardless of geography, but am open to suggestions... (or maybe it's OK if we don't mention it!)

distalx:
@lpatmo Thanks for your feedback. 
**RE : striped table**
I was wondering that for `member` section why did we switch to `table`? but this clears it now.
Striped table look is also feasible with  the rows, it will need some CSS magic there. I think it's easier to manage resposive views with row and cols. let me know your thoughts.

**RE : clarify that time conversions**
maybe we could use dismissive alerts!?

I'm sure that we will get more feedback from others once we tested into staging.

client/templates/availability/availability_slot_local.js
```diff
@@ -0,0 +1,11 @@
+import UHR from 'uhr';
+
+Template.availabilitySlotLocal.helpers({
+  localSlotTime: function(){
+    const slot = Template.instance().data;
+    let result = UHR(slot.d, slot.h, slot.m, -1)
+    result.hour = result.hour < 10  ? `0${result.hour}` : result.hour;
+    result.minute = result.minute < 10  ? `0${result.minute}` : result.minute;
+    return result;
```
nalbina: shouldn't this function return a time rather than a result?

client/templates/study_groups/study_group_availability.html
```diff
@@ -0,0 +1,88 @@
...
+
+
+  {{#if Template.subscriptionsReady}}
+
+    <div class="row">
+      <div class="col-md-6">
+        <div class="page-header">
+          <h3>Members' Availability</h3>
+        </div>
+        {{#each groupMemebersAvailabilities}}
```
nalbina: typo

client/templates/study_groups/study_group_availability.js
```diff
@@ -0,0 +1,36 @@
+Template.studyGroupAvailability.onCreated(function() {
+  let instance = this;
+  let studyGroupId = FlowRouter.getParam('studyGroupId');
+  instance.subscribe('allStudyGroupAvailability', studyGroupId);
+
+});
+
+Template.studyGroupAvailability.helpers({
+  groupMemebersAvailabilities (){
```
nalbina: typo

client/templates/study_groups/edit_study_group_availability_slot_modal.html
```diff
@@ -0,0 +1,128 @@
+<template name="editStudyGroupAvailabilitySlotModal">
```
nalbina: maybe we could find a datepicker component for this :)

client/templates/availability/availability_slot_local.js
```diff
@@ -0,0 +1,11 @@
+import UHR from 'uhr';
```
nalbina: could we also use momentjs for this?

client/templates/availability/availability_slot_local.js
osanseviero: The result of the function is an UHR, which is a time object. 
Something like: { day:Number, hour:Number, minute:Number }

Here is the package: https://libraries.io/npm/uhr/0.0.2

nalbina: then the variable could be `timeObject`

client/templates/study_groups/edit_study_group_availability_slot_modal.html
distalx: done 👍

client/templates/availability/availability_slot_local.js
distalx: `moment` is a Swiss Army Knife for dealing with Date-Time in JavaScript, but still it's a headache to deal with the `Date` object.
`uhr` is specifically design for this purpose only. and has a simple api.
when it comes to performance `uhr` is more efficient than `moment`.
I have setup a benchmark repo here.
https://github.com/distalx/uhr-benchmark

nalbina: I see you have done your research already. lets go with uhr :)

PR #719, opened 2017-11-18
Issue 691 hangout participant count

### LINKED ISSUE #691: See number of people in 24/7 Hangouts
When going in the Study Groups page, it would be great to see how many people are actually online in the 24/7 Hangout.

At the moment there is no way to know if someone is there beside connecting into it. So if we're looking for company, we would potentially have to connect in every single Hangout.

I believe those 24/7 Hangouts are not used so much for that reason. There is a lack of visibility on who is currently online in the Hangout.

railsstudent:
Fixes #691.

Create a MongoDB collection app_stats with schema  { hangout_id: string, participant_count: number }
A unique index should be created for hangout_id 
// add unique index, db.app_stats.createIndex({ hangout_id, 1}, { unique: true });

Server publishes hangoutParticipants and allHangoutParticipants to return number of participants in a specific hangout room

In single study group.js, participant count is shown in a badge in 24/7 hangout tab
In hangout_frame.js, participant count is shown in a well below jitsu hangout.  participant count is incremented when user loads jitsu and decremented when user leaves the page
In all study room.js and my study room.js, participant count of 24/7 hangout is listed for each study room; adjacent to laptop icon


railsstudent:
all_study_groups_participant_count

railsstudent:
@distalx  Changes made to get rid of the session variable

server/hangouts/methods.js
```diff
@@ -404,3 +404,36 @@ Meteor.methods({
 
   }
 });
+
+Meteor.methods({
```
distalx: Please Move methods related to `AppStats` collection under a  `/server/app_stats/methods.js` directory.
`joinParticipant` and `leaveParticipant`

server/hangouts/publications.js
```diff
@@ -82,3 +82,22 @@ Meteor.publish( 'hangoutSearch', function(searchTerm) {
 
     return Hangouts.search(searchTerm);
 });
+
+Meteor.publish('hangoutParticipants', function(room) {
```
distalx: `hangoutParticipants` and `allHangoutParticipants` These publications should be also placed under `/server/app_stats/publications.js` directory.

server/hangouts/publications.js
```diff
@@ -82,3 +82,22 @@ Meteor.publish( 'hangoutSearch', function(searchTerm) {
...
+  check(room, String);
+  if (room) {
+      return AppStats.find({ hangout_id: room });
+  }
+  return this.ready();
+});
+
+Meteor.publish('allHangoutParticipants', function(hangoutRoomIds) {
+  if (hangoutRoomIds == null || typeof hangoutRoomIds === 'undefined' ||
```
distalx: Please make use of `check` here.
i.e. check(hangoutRoomIds, Match.Maybe([String]))

server/hangouts/methods.js
railsstudent: Fixed

server/hangouts/publications.js
railsstudent: Fixed

server/hangouts/publications.js
railsstudent: Fixed

railsstudent:
@distalx  Changes made

railsstudent:
@distalx  Please kindly review this PR when you have time. Thanks.

distalx:
Great work @railsstudent! 👏 . Love seeing it all come together! LGTM! 👍

I have made few changes.
- storing a participant object inside participants list instead of just a `count` to avoid duplicates.
- adding a user to participants list when user join the hangout by using a `join here link`.

**_ Although there is a one scenario which is really hard to get around._**

When a person join the hangout by using a `join here link` it will open the `Jitsi` in a _new tab_. We are not able to remove that person from the list of participants when that person close down the _`Jitsi` tab_. We are only able remove that person from the list of participants when that person either close down the `CB app` or _log off_ from it.

So there are two possibilities, when user won’t get the accurate participants count.

1) when a person close down the `Jitsi` tab and don’t close down the `CB app` or _log off_ from it.
2) when a person close down the `CB app` or _log off_ from it and don’t close down the `Jitsi` tab.

Possible solutions !

1) remove the `join here link`.
	it reduces the complexity of tracking user’s actions out side the `CB app`. we have added the `join here` link because  sometimes `Jitsi` fails during loading process and ppl might have to refresh the page to start the load process again.

2) run a cron job.
          as a last resort we could always implement a cron job which trigger at every couple of hours and reset/update the active participant count. it isn’t directly addresses the issue on hand but it reduces the possibility of stale data which is the by product of our current implementation.


cc @lpatmo , @sergeant-q.

lpatmo:
Thanks for the hard work here, @railsstudent and @distalx! 👏 

> When a person join the hangout by using a join here link it will open the Jitsi in a new tab. We are not able to remove that person from the list of participants when that person close down the Jitsi tab. We are only able remove that person from the list of participants when that person either close down the CB app or log off from it.

I'm not too worried about this scenario, actually. If they close the Jitsi tab but forget to close the CB app, they'll still technically be in the hangout, right? Of the two proposed solutions, I would lean towards the cron job instead of removing the "click here" link, since I think the user experience is more terrible if you can't get in to the hangout at all. Plus, the external link gives us easy access to the URL for the Jitsi room.

Shall I punt the cron job into a new issue -- and in the meantime, merge to staging?

lpatmo:
One thought: in `client/templates/study_groups/all_study_groups.html` and in `client/templates/study_groups/my_study_groups.html`, do you think it might make sense to only show the laptop icon with the number count next to it if there is at least one person in the 24/7 room?

screen shot 2017-12-09 at 10 35 53 pm

I.e.
``` 
+ {{# if numParticipants _id }}<small><i class="fa fa-laptop fa-fw"></i>{{ numParticipants _id }}</small>{{/if}}
- <small><i class="fa fa-laptop fa-fw"></i>{{ numParticipants _id }}</small>
```

railsstudent:
@lpatmo  Okay. I will work on it in the weekend.  
I have slacked off since taking an interest in golang

lpatmo:
@railsstudent If you feel like *you've* slacked off, then I've been... comatose? :P 

The change is pretty straightforward -- it would be those two lines I pasted above for  `all_study_groups.html` and `my_study_groups.html`, but before we make it let me know what you think of the idea! I'm not against showing it at all times either; am mostly worried that potentially having a lot of 0s will be off-putting.

railsstudent:
@lpatmo I think it is a good idea to hide the laptop icon when no one is in 24/7 hangout.
Since the study group card is so small, it is prudent to convey only meaningful data.

lpatmo:
Thanks!

railsstudent:
@lpatmo  Templates fixed. Please kindly review

PR #730, opened 2017-11-19
Issue 446 Adds language detection for TAPi18n

### LINKED ISSUE #446: Detect user language to take advantage of i18n translations
@osanseviero added an es.i18n.json file, which means we now have the capacity to translate to Spanish. 

According to https://github.com/TAPevents/tap-i18n#quickstart, we need to detect the language used by the user:
```
getUserLanguage = function () {
  // Put here the logic for determining the user language

  return "sp";
};

if (Meteor.isClient) {
  Meteor.startup(function () {
    Session.set("showLoadingIndicator", true);

    TAPi18n.setLanguage(getUserLanguage())
      .done(function () {
        Session.set("showLoadingIndicator", false);
      })
      .fail(function (error_message) {
        // Handle the situation
        console.log(error_message);
      });
  });
}
```

Will research later how to do this.

ajparise:
Fixes #446.

Took inspiration from https://themeteorchef.com/tutorials/i18n-and-meteor#tmc-auto-setting.

Default language fall back is 'en'.
Language can forced using the query param `lang`.

### DIFF
--- a/client/templates/main.js
+++ b/client/templates/main.js
@@ -9,6 +9,12 @@ if (Meteor.isClient) {
       console = console || {};
       console.log = function(){};
     }
+
+    const defaultLang = 'en'
+    const urlLang = FlowRouter.getQueryParam('lang')
+    const browserLang = (window.navigator.userLanguage || window.navigator.language || '').slice(0,2)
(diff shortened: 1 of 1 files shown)


nalbina:
Works like a charm. Great job, @ajparise. LGTM

PR #731, opened 2017-11-19
Issue 592 transfer ownership v2

### LINKED ISSUE #592: STUDY GROUPS: Let owners transfer ownership of study group to another user
Not sure if this should be a button in the "settings" tab, or if owners should automatically be able to assign "owner" in the members' tab when they click to modify the permissions of other users.

Thoughts?

railsstudent:
Fixes #592 

A new section TRANSFER STUDY GROUP is visible in Settings tab if the group as at least 1 non-owner member
When owner transfer study group to another member, the current owner becomes member of the study group and a new activity is logged to Activity tab
Afterward, old owner is redirected to all study group

transfer_study_group_setting
confirmation_dialog
new_member_status
study_group_activities


nalbina:
hey @railsstudent is this WIP?

railsstudent:
it is completer

railsstudent:
it is completed

nalbina:
I am somehow missing the option to transfer the group to another member. are you sure you added the relevant html portion or could you point me to it? :)

railsstudent:
in settings tab.
must be owner and have more than 1 member in the group

railsstudent:
in settings tab.
must be owner and have more than 1 member in the group

railsstudent:
U are right, when i make a new branch and squash commit, i lost the changes in settingd. I will add it back tonight

railsstudent:
@nalbina   Fixed.   Please kindly review.
Amendment: After study group is transferred, member remains in the study group and Description tab becomes visible.

client/templates/study_groups/study_group_settings.js
```diff
@@ -89,8 +103,9 @@ Template.studyGroupSettings.events({
             return Bert.alert( error.reason, 'danger', 'growl-top-right' );
           }
           if(result){
-            FlowRouter.go("all study groups");
-            return Bert.alert( 'Study Group transferred', 'success', 'growl-top-right' );
+            // Select the first tab in the study group
+            $('div.study-group-body > ul li:first-child > a').tab('show');
```
angelocordon: ` $('div.study-group-body > ul li:first-child > a')` seems a little crazy - can it be cleaned up for better readability?

Probably adding a `.study-group-tab` class on those links and querying it that way?

```javascript
$('.study-group-tab').first().show();
```

railsstudent: @angelocordon  Your expression does not work.  
I change it to a more readable one.

ajparise: Here's a possible middle ground between complexity and hard coding:

```javascript
$('div.study-group-body [role=tablist] a:first').tab('show')
```

angelocordon: @railsstudent curious to hear why it didn't work? It may require a couple more lines of code, but in the end, I think pushing for code readability will level up our code quality and help other contributors be able to understand our codebase better and raise our standards (which is always a good thing!). 

I'm not familiar with how we're using `.tab()` so that may be it - 

If you add class names for the links you can use that as selectors directly.

```html
<div class="study-group-body">
  <ul>
    <li>
       <a class="study-group-tab">...</a>
       <a class="study-group-tab">...</a>
         ...
</div>
```

```javascript
$('.study-group-tab').first().tab('show')
```

The HTML snippet above is a bit of a pseudo code just to show what I'm talking about. If this template is in the `staging` branch I can dig deeper. 

In any case, what do you guys think? Interested in hearing your thoughts - does this make it clearer/sense?

cc: @ajparise

ajparise: @angelocordon I'm of the opinion adding class selectors makes it more confusing and harder to maintain. Those classes serve no purpose other than reducing the complexity of a selector.

`$('div.study-group-body > ul li:first-child > a')` is precise and only selects exactly what it's supposed to. I think that actually improves the readability.

server/study_groups/methods.js
```diff
@@ -454,3 +454,59 @@ Meteor.methods({
...
+      studyGroupTitle: String,
+      studyGroupSlug: String
+    });
+
+    if (!this.userId) {
+      throw new Meteor.Error('Members.methods.editStatus.not-logged-in', 'Must be logged in to transfer study group.');
+    }
+
+    const actor = Meteor.user()
+    if (!actor || !Roles.userIsInRole(actor, ['owner','admin', 'moderator', 'member'], data.studyGroupId )) {
```
distalx: Please modify this.
Only owner should be able transfer the membership. 
i.e. `!Roles.userIsInRole(actor, ['owner'], data.studyGroupId )`

server/study_groups/methods.js
```diff
@@ -454,3 +454,59 @@ Meteor.methods({
...
+      throw new Meteor.Error(403, "Access denied");
+    }
+
```
distalx: Here, we should also check if the `newOwnerId` is belong to any of the member or not.

`if(!Roles.userIsInRole(data.newOwnerId, ['admin', 'moderator', 'member'], data.studyGroupId ){`

server/study_groups/methods.js
```diff
@@ -454,3 +454,59 @@ Meteor.methods({
...
+    //Remove old owner and update role of new owner to 'owner'
+    StudyGroups.update({_id: data.studyGroupId, 'members.id': actor._id },
+                     {$set: {'members.$.role': 'member'} });
+    StudyGroups.update({_id: data.studyGroupId, 'members.id': data.newOwnerId },
+                     {$set: {'members.$.role': 'owner'} });
+    Roles.setUserRoles(actor._id, 'member', data.studyGroupId);
```
distalx: Instead of demoting current `owner` to a `member` role, what if we just make them an `admin` ?

client/templates/study_groups/study_group_settings.html
```diff
@@ -23,5 +23,20 @@ <h3 class="panel-title">Archive this study group</h3>
     </div>
   </div>
 
+  {{#if eligibleMembers.length }}
+  <div class="panel panel-default">
+    <div class="panel-heading">
+      <h3 class="panel-title">Transfer Study Group</h3>
+    </div>
+    <div class="panel-body">
+      <select class="form-control" id="studyGroupMemberList">
```
distalx: Maybe we could use a `select2` component for this :)

client/templates/study_groups/study_group_settings.js
railsstudent: Sad to hear my code quality is not good.  
I will take a step back to focus on learning of web development and let others tackle the issues.

angelocordon: @ajparise I hear what you're saying, but I don't fully agree 😃 

I'm open to your opinions, however I think adding the class names to those links does exactly what you're saying - if we query the links via `$('.study-group-tab')`, jQuery traverses the DOM specifically for those elements and only those elements. 

When we pass `div.study-group-body > ul li:first-child > a` as a query selector, it works from right to left; it will first collect all `a` elements, then it will find the `a` elements within a `li` then `ul`, and then scope it down to `div.study-group-body`. 

Vs when we tell it to look specifically for an element with a specific class name, then it only needs to traverse the DOM once 😄 

Re: readability, perhaps `study-group-tab` does not have the correct semantics; `nav-tab` may be better? I'd argue that `$('.nav-tab').first()` is definitely more understandable. 

Alternatively, looks like we can also use `a[role="tab"]`. There are some slight performance differences regarding querying elements with classes vs attributes, but I guess they don't matter as much in 2017. 

```javascript
$('a[role="tab"]').first().tab('show')
```

or 

```javascript
// requires you to add a `nav-tab` class to the anchor links
$('.nav-tab').first().tab('show')
```

Either of the above does what you want @railsstudent.

angelocordon: @railsstudent - please don't feel discouraged by all of this, @distalx gave some great feedback and this discussion between @ajparise and I is interesting in a sense that I hope you can find some insights from all of our perspectives. 

It's perfectly okay to take a step back, but also do it so you can get a better scope of the feature you're working on :)

distalx:
Great work @railsstudent 👏 

thanks for the review @angelocordon , @nalbina.  💯

PR #735, opened 2017-11-22
Language switching and local storage persistence

ajparise:
Supplements #446, #730.
- Adds language switching to footer
image
- Language list is generated dynamically from `TAPi18n.getLanguages()`
- Persisted via LocalStorage
- Removes query param for forcing language

image
image

### DIFF
--- a/client/templates/home/footer.js
+++ b/client/templates/home/footer.js
@@ -19,6 +19,8 @@ Template.footer.events({
     event.preventDefault();
     const languageCode = event.target.dataset.langCode
     const languageName = event.target.dataset.langName
+    if (languageCode === TAPi18n.getLanguage())
+      return;
 
     TAPi18n.setLanguage(languageCode)
     .done(() => {
--- a/client/templates/home/footer.js
+++ b/client/templates/home/footer.js
@@ -29,7 +29,7 @@ Template.footer.events({
       console.log(error);
     })
     .always(() => {
-      localStorage.setItem('lang', TAPi18n.getLanguage());
+      localStorage.setItem('languageCode', TAPi18n.getLanguage());
     })
   }
 });
--- a/client/templates/home/footer.js
+++ b/client/templates/home/footer.js
@@ -1,0 +1,35 @@
+Template.footer.helpers({
+  languageSwitchers: function() {
+    const supportedLanguages = TAPi18n.getLanguages();
+    const currentLanguageCode = TAPi18n.getLanguage();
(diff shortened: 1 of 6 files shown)


nalbina:
Hey @ajparise, thanks for your PR. Does it need a review yet?

ajparise:
@nalbina It's ready for a review. Thanks!

client/templates/home/footer.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+});
+
+Template.footer.events({
+  "click [data-lang]": function(event, template){
+    event.preventDefault();
+    const languageCode = event.target.dataset.langCode
+    const languageName = event.target.dataset.langName
+
+    TAPi18n.setLanguage(languageCode)
```
nalbina: I feel like setting the same language again should be a noop. right now I am getting a notification for each click.

client/templates/home/footer.js
```diff
@@ -0,0 +1,35 @@
...
+    .done(() => {
+      Bert.alert(TAPi18n.__("alert_language_change", `${languageName} (${languageCode})`), 'success', 'growl-top-right' );
+    })
+    .fail((error) => {
+      Bert.alert(TAPi18n.__("alert_language_change_fail", `${languageName} (${languageCode})`), 'danger', 'growl-top-right' );
+      console.log(error);
+    })
+    .always(() => {
+      localStorage.setItem('lang', TAPi18n.getLanguage());
```
nalbina: Can we call the localstorage item also languageCode to stay in sync with the rest? :)

client/templates/main.js
```diff
@@ -9,14 +9,14 @@ if (Meteor.isClient) {
     }
 
     const defaultLang = 'en'
-    const urlLang = FlowRouter.getQueryParam('lang')
+    const localStorageLang = localStorage.getItem('lang');
```
nalbina: Can we call the localstorage item also languageCode to stay in sync with the rest? :)

client/templates/home/footer.js
ajparise: Makes sense and was intentional, but valid point. I added a guard clause.

client/templates/home/footer.js
ajparise: Good catch, updated in related files.

distalx:
Great work @ajparise 👏
thanks for the review @nalbina 🥇

PR #736, opened 2017-11-22
Issue 732 study group prompt login

### LINKED ISSUE #732: "Must be logged in to create StudyGroup"
I feel like this is a huge bummer when I already thought out my studygroup title, description and tags only to see that message when I try to create a study group. Maybe we should blank out the possibility to do that entirely while logged out or prompt them to login.

railsstudent:
Fixes #732.

If user is not login, reuse the Sign in with Slack dialog in layout.html.


client/templates/study_groups/all_study_groups.html
```diff
@@ -7,7 +7,7 @@
           <div class="row">
               <div class="col-md-12 align-center">
                 <h2>Study Groups</h2>
-                <div class="btn btn-cb2" id="createGroupButton"> {{_ "start_new_group"}} </div>
+                <div class="btn btn-cb2 {{#unless currentUser}} continue-popup {{/unless}}" id="createGroupButton"> {{_ "start_new_group"}} </div>
```
distalx: Please add this `id="{{#if currentUser}}createGroupButton{{/if}}"`. and we won't be needing a 
`if (Meteor.userId()) {` block in our helper function.
Thanks.

railsstudent:
@distalx  if button has no id, the CSS styling of create study group will not be applied. Therefore, I introduce a btn-createGroup-NotLoggedIn class and apply the same styling.

railsstudent:
@distalx  Fixed. Please kindly review. Thanks.

distalx:
@railsstudent Thanks for the PR!

PR #743, opened 2017-12-17
Schedule hangout default studygroup name

railsstudent:
When I schedule new hangout in single study group, study group name is not default in the new hangout modal. I have to type study group name in the drop down to narrow down the selections and select the group name manually.

Similarly, it also occurs when duplicating hangout in hangout page and single study group. This change will populate study group name in addition to description and topic


client/templates/hangout/hangout-clone.js
```diff
@@ -53,6 +53,10 @@ Template.cloneHangoutModal.onRendered(function() {
         data: studyGroups
       });
 
+      const groupId = templateInstance.data.hangout.data.group.id || '';
+      if (groupId !== '') {
+    -   instance.$(".study-group-single").val(groupId).trigger('change');
```
lpatmo: This is awesome, @railsstudent. Thanks for addressing this issue.

Question: Was this `-` on line 58 added by mistake, or am I missing something?

railsstudent: It is a typo. Added by mistake.

railsstudent: I will make the correction tonight

railsstudent: @lpatmo  Fixed. Please kindly review

PR #745, opened 2017-12-19
Fix account deletion.

distalx:
Account deletion functionality is broken because of a typo. This PR fixes it.
H/t to **Void** for pointing out in slack.


distalx:
@railsstudent thanks for the review.

PR #760, opened 2018-01-06
Refactor about page v2

railsstudent:
Externalize details of contributors in private/about.json and read the data from it when rendering the images and popups in about.html.

When new contributors wants to add themselves to about page, they can append JSON object of personal details in code-contributors array.   

The html code in title attribute in original design is errone hone with all these angle brackets


railsstudent:
@lpatmo  Yes, you can move my image to core contributor session.
My slack handler name is @Connie
twitter is con_leung
github is railsstudent

Thanks.

private/about.json
```diff
@@ -0,0 +1,235 @@
+{
```
distalx: We should move this file under `imports/data` directory, and break it down to two files, 
The scope of the private directory is limited to server side only.

very roughly : 

`imports/data/flags.json` and `imports/data/contributors.json`
```
[
  {
     name: 'xyz',
     avatar: 'http://link or gravater',
     intro: 'lorem ipsum',
     socials:[
         {website: 'URL'},
         {twitter: '@handel'},
     ....
     ],
     type: 'code'
  },
   ....
]
```

on about page you can load this `json` files with their relative path
```
imports contributors from 'imports/data/contributors.json'
```
with the help of the helper function contributors can be passed to the template 
```
`contributors : function() { return contributors }`
```
and then we can iterate over it using 
`{{#each contributors}} {{/each}}` block.

To display contributors into their respective section we can use `equals` template helper  `{{#if equals type 'code'}}`

railsstudent: I do not see imports/data directory in the project. Should I create it? thanks.

distalx: Yes, `imports` directory is a part of a new directory structure suggested in meteor guide, our migration to the new directory structure is long over due.

railsstudent: Noted. Will get this done before Sunday.

railsstudent: @distalx  Please kindly review

distalx:
@railsstudent , Thanks for the pull request.

PR #778, opened 2018-01-27
Welcome page template, copy, route

### LINKED ISSUE #669: Product tour of study groups feature (and platform in general)
UPDATE: renamed issue!

When I head over to a study group page (e.g. https://codebuddies.org/study-group/tdd-with-python/AshYgJsHdyo6kuCQW), the first tab I want to click on is "Members," because it'll show the statuses of each study group member.

screen shot 2017-10-15 at 5 50 00 pm

I know to do this because I'm familiar with the feature, but newcomers might never know or think to click on the "Members" tab, and study group members themselves might not know that they can input their current status.

I'm tempted to remove the "Members" tab and move its content below the Description.

What does everyone think? Or does anyone have alternate suggestions for highlighting the feature found in "Members"?

julianjohannesen:
Fixes #669

Created the route (lib/routes.js) and template (in client/templates/other) for a new "Welcome" page  and added the content found in the new #start-here Slack channel.

### DIFF
--- a/client/templates/other/welcome.html
+++ b/client/templates/other/welcome.html
@@ -1,0 +1,71 @@
+<template name="welcome">
+  <div>
+    <section class="conatainer-fluid section-header">
+      <h1>Welcome Guide</h1>
+      <a href="{{pathFor 'all study groups'}}" class="btn btn-link">
+        <i class="fa fa-arrow-left"></i> Back to Study Groups
+      </a>
+    </section>
+
+    <section class="section">
+      <div class="container">
+        <div class="row">
+          <div class="col-sm-12 col-md-8 col-md-offset-2 section-body">
+            <p>
+              We learn better together.
+            </p>
+            <p>
+              We’re a community of independent code learners sharing knowledge & helping each other learn faster. We come from all over
+              the world. Everyone is welcome - no previous experience required.
+            </p>
+            <p>
+              CodeBuddies is a 100% volunteer-run community & our website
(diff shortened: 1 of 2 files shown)


client/templates/other/welcome.html
```diff
@@ -0,0 +1,71 @@
+<template name="welcome">
+  <div>
+    <section class="conatainer-fluid section-header">
+      <h1>Welcome Guide</h1>
+      <a href="{{pathFor 'all study groups'}}" class="btn btn-link">
```
angelocordon: 50/50 here but will leave it to you @julianjohannesen - 

Should this link back to `/hangouts` as that's the page that you land in after a user signs in? If so, we should change the text to "Start or Join a Hangout"

Likewise, we should change the text to "Start or Join a Study Group" to encourage more actionable feedback to the user.

angelocordon:
Looks good @julianjohannesen - just left a single comment but not blocking for a review or anything 👍

client/templates/other/welcome.html
julianjohannesen: Sorry about the delayed reply. From what Linda was telling me, it sounded like there was a longer conversation about where it should link to. Not sure what makes the most sense. I like the "start or join" language.

lpatmo: I like the "start or join" language too! Will merge this in for now, and we can make the language tweaks in a new issue/PR.

PR #796, opened 2018-02-17
Issue 581 add Join/Leave study group in study group pages

### LINKED ISSUE #581: Study Groups: Join/Leave from "My Study Groups" Page
A drop-down menu on the listings where you could join/leave shown on "My Study Groups" page would be nice.

railsstudent:
Fixes #581.

In All study groups page, Join this group button is rendered if user is not member/admin/moderator of a study group. 
If user is member/admin/moderator of a study group, Leave this group button is rendered and he/she can click it to leave the group without having to navigate to study group details page.

Similarly, Leave this group button is rendered in My study groups page to allow users to leave the group without navigation to study group details page.

all_study_group_join_leave_button
my_study_groups_leave_button


lpatmo:
Thank you for adding this feature, @railsstudent! 🙏 

Only jarring user experience that I see is that the study group card moves to the end of the list whenever I click to leave or join one. 

I think the problem is caused by the fact that the study groups are being loaded and sorted by a `created_at` parameter, but `created_at` does not exist in the schema.

So I think this may help:
```
   instance.loadStudyGroups = function() {
-    return StudyGroups.find({}, {sort: {createdAt: 1}});
+    return StudyGroups.find({}, {sort: {title: 1}});
   }
```
(from `client/templates/study_groups/all_study_groups.js`)

Let me know what you think, Connie! Thanks.

railsstudent:
@lpatmo  Linda, it is fixed, thanks.

lpatmo:
@railsstudent Thank you!!

PR #797, opened 2018-02-18
Edit typo at Line 35

rmorabia:
Forgot the space between ## and the start of the heading.

Not sure if I should've opened an issue first? It seemed kind of small to go through the entire process.


lpatmo:
@rmorabia Thanks for this PR! Agreed, didn't really need an issue. :)

PR #776, opened 2018-01-27
Manage email preference

distalx:
This PR is a part of an issue #626 

User can manage email notifications.

screen shot 2018-01-27 at 7 55 03 pm

Issue #626, opened 2017-09-27
Email notifications: alert when new member joins a study group


lpatmo:
Send an (optional) email notification to owner/admins/moderators of a group when a new member joins. 

Note: owners/admins/moderators should be able to customize their email subscriptions within each group.

PR #801, opened 2018-03-29
Email notifications

distalx:
This PR is a part of an issue #626 and follows #776 

Subscribers will be notified on following events
- new hangouts 
   when user creates a hangout in a study group.
- new member
   when new members joins the group.
- new discussion 
   when user creates a discussion in a study group.

User can set notification preference from their account settings.

A cron job riggers these notifications.
Right now cron triggers at every 40 minute but for production we might want to tune it up to 20!.

### DIFF
--- a/imports/libs/server/cb_mailer/index.js
+++ b/imports/libs/server/cb_mailer/index.js
@@ -1,0 +1,33 @@
+import { Meteor } from 'meteor/meteor';
+import {SSR} from 'meteor/meteorhacks:ssr';
+import { Email } from 'meteor/email';
+
+/**
+* CBNotifier globle notifier.
+* @function
+* @name CBMailer
+* @param { Object } mail_data - sender receiver information
+* @param { String } template_name - which template to use
+* @param { Object } template_data - template_data
+*/
+export const CBMailer = function(mail_data, template_name, template_data) {
+
+  Meteor.settings.public.isModeDebug ? console.log(`inside mailer ${template_name}`) : null;
+
+  SSR.compileTemplate(`${template_name}`, Assets.getText(`templates/email/${template_name}.html`));
+
+  const mail = {
+    to: mail_data.to,
+    from: mail_data.from,
+    subject: mail_data.subject,
+    html: SSR.render(template_name, template_data)
+  }
+
+  try {
+    Email.send(mail);
+  } catch (e) {
+    console.log(e);
+
+  }
+
+}
--- a/imports/libs/server/cb_mailer/index.js
+++ b/imports/libs/server/cb_mailer/index.js
@@ -1,5 +1,5 @@
 import { Meteor } from 'meteor/meteor';
-import {SSR} from 'meteor/meteorhacks:ssr';
+import { SSR } from 'meteor/meteorhacks:ssr';
 import { Email } from 'meteor/email';
 
 /**
@@ -27,7 +27,6 @@ export const CBMailer = function(mail_data, template_name, template_data) {
     Email.send(mail);
   } catch (e) {
     console.log(e);
-
   }
 
 }
--- a/imports/libs/server/cron/helpers.js
+++ b/imports/libs/server/cron/helpers.js
@@ -1,0 +1,175 @@
+import moment from 'moment';
+
+/**
+* get list of users who are eligible to receive emails for given type
+* @function
+* @name getEligibleRecipients
+* @param { String } type - email notification type
+* @param { Array } subscribers - list of user(s)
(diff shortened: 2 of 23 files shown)


server/hangouts/methods.js
```diff
@@ -88,6 +88,11 @@ Meteor.methods({
         id: group._id,
         title: group.title,
         slug: group.slug
+      },
+      email_notifications: {
+        initial: false,
+        reminder: false,
+        follow_up: false
```
lpatmo: Does this mean users of a group won't receive email alerts when a hangout in that group is created?

imports/libs/server/cron/initial_email_notifications.js
```diff
@@ -0,0 +1,268 @@
...
+               }
+
+              CBMailer(mail_data, template_name, template_data);
+            })
+
+
+            // set intial flag for email_notifications true
+            listOfNewHangouts.forEach(function(hangout){
+              Hangouts.update({_id:hangout._id}, {$set:{
+                'email_notifications.initial': true
```
lpatmo: .... oh wait, it's set to true here!

lpatmo:
Looks beautiful.

One bug I noticed: if you start a new discussion from the nav bar (instead of from inside a study group), the attempt fails because the discussion needs to be linked with a groupId. 

Maybe the form for a new discussion started on the nav header needs to be different -- or include the list of groups?

Or perhaps we can remove the "Start a new discussion" button for now, and tell users to start new discussions from inside their groups.

imports/libs/server/cb_mailer/index.js
```diff
@@ -0,0 +1,33 @@
+import { Meteor } from 'meteor/meteor';
+import {SSR} from 'meteor/meteorhacks:ssr';
```
nalbina: whitespace

imports/libs/server/cb_mailer/index.js
```diff
@@ -0,0 +1,33 @@
...
+    from: mail_data.from,
+    subject: mail_data.subject,
+    html: SSR.render(template_name, template_data)
+  }
+
+  try {
+    Email.send(mail);
+  } catch (e) {
+    console.log(e);
+
```
nalbina: whitespace

imports/libs/server/cron/helpers.js
```diff
@@ -0,0 +1,175 @@
...
+/**
+* get list of users who are member of a group
+* @function
+* @name getGroupMembers
+* @param { String } type - email notification type
+* @param { Array } subscribers - list of user(s)
+* @return { Array } members - list of group members
+*/
+function getGroupMembers(study_group_id) {
+  if (study_group_id === 'CB') {
```
nalbina: ?

imports/libs/server/cron/initial_email_notifications.js
```diff
@@ -0,0 +1,268 @@
...
+  createdFromNow } from './helpers';
+
+
+/**
+* send email notification for new hangouts to eligible recipients
+* @function
+* @name initialEmailNotificationsForHangouts
+*/
+async function initialEmailNotificationsForHangouts() {
+   let hangouts = await getNewHangouts();
```
nalbina: maybe this is worth wrapping in try catch.

server/hangouts/methods.js
distalx: No, These flags are there so cron job can keep track for which notification has been sent.
we makes it true once notification is sent. [here](https://github.com/codebuddies/codebuddies/pull/801/files/8d65e59ddf86bfbc9e7a295b7252a85827ef7ff9#diff-6c4f4a0e79a2047c67141624ce02c65aR73)

distalx:
@lpatmo  Thank you for reviewing this.

>One bug I noticed: if you start a new discussion from the nav bar (instead of from inside a study group), the attempt fails because the discussion needs to be linked with a groupId.

In that case Discussion is linked to the `CB`(global) group.

>Maybe the form for a new discussion started on the nav header needs to be different -- or include the list of groups? Or perhaps we can remove the "Start a new discussion" button for now, and tell users to start new discussions from inside their groups.

The idea behind this behaviour is that user can start a discussion without being a part of any group. same as a hangout. and yes we should include the list of groups.

imports/libs/server/cron/initial_email_notifications.js
distalx: `getNewHangouts` doesn't throw any err, So I'm not sure what purpose would it serve.

PR #804, opened 2018-04-02
Popup break word

railsstudent:
When mouseover a hangout card, a popup is placed on top with content. I discovered that long URL overflow the popup unexpectedly. I added CSS word-wrap property to break word such that every word and URL is displayed within the boundary of the popup.


railsstudent:
overflow-text-in-popup

railsstudent:
break-word-2

PR #805, opened 2018-04-03
Encode google calendar topic/description

railsstudent:
If hangout topic/description contains ampersand character (&), the text is truncated before the first occurrence of & and displayed in google calendar form.

EncodeURIComponent is applied to topic and description respectively before  the values are append to the url of anchor element. When google calendar form is opened in new tab, full topic and description are displayed properly. 

incomplete_google_calendar
complete_google_calendar


lpatmo:
LGTM. Thank you!

Issue #457, opened 2017-02-17
Add support for logging in using Github authentication


lpatmo:
Slack sign-in often confuses users, because if you're not logged into any Slack account, you have to remember to type in `codebuddies`.slack.com and then remember your login to that Slack account. 

If we add support for Github authentication, we'll have to make sure to at least get the following fields from the Github API:
- user's email
- user's username
- user's avatar image URL

newswim:
I really like this idea, it's been confusing for myself and others in the past. ++

sergeant-q:
@lpatmo I might have time to take this feature on. Any other specific requirements for user info and are we adding this auth strategy in addition to Slack?

lpatmo:
Cool! Thinking about it a little more, the minimum info. we need from Github are:
- username
- email
- avatar image ("image_192")

And yeah, we're adding this as an auth strategy in addition to Slack, not replacing Slack.

One possibly tricky thing that just occurred to me: `username` assumes a Slack username aka

```
{{# if userInfo.username }}
                  <a href="https://codebuddies.slack.com/messages/{{userInfo.username}}/" target="_blank">
                    <i class="fa fa-slack" aria-hidden="true"></i>
                  </a>
                {{/if}}
``` // profile.html

... so we may have to edit the logic to check for a `userInfo.slack` and make sure that we have a data key named `slack` in our mongo record. 

And we may need to double check that we're not conflating 'username' with Slack usernames elsewhere, as well.

If that doesn't make sense, let me know! Happy to also file a separate issue about it to separate out the steps.

lpatmo:
Backing up a step:
### Disadvantages of supporting Github signin:
- Slack signin usernames will be lost, AKA no way to tie it to a Github login 
- Would be hard to tell who is who on the site, since much of the community/activity exists on Slack (but then again: what if the community moved to the site? Or to a Discourse forum with Github login?)

### Advantages of supporting Github signin:
- easier signin compared to the two-or-three-step process required with Slack

===
Or maybe we simply need clearer instructions about how to log in using Slack? (That is its separate issue already in https://github.com/codebuddiesdotorg/codebuddies/issues/466).

nalbina:
Can't we offer both options at the same time? I know that we could use accounts-github for the github stuff and leave slack like it is?

lpatmo:
[...]

PR #806, opened 2018-04-08
Signin workflow improvement

distalx:
This PR is a part of an issue #457 and follows #802

 - If username is already in use, set a new username.
 - on first time sign in, let user set their basic information (username, firstname, lastname)

### DIFF
--- a/client/templates/other/has_blocked.js
+++ b/client/templates/other/has_blocked.js
@@ -1,11 +1,19 @@
 "use strict"
 
+import '../../../imports/ui/pages/users/basic_user_information';
+
 Template.hasBlocked.helpers({
   target: function () {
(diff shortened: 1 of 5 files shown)


lpatmo:
Thanks @distalx. After I checked out the branch, I would get logged out after signing in with Github:

1) Signed in with Github; looked okay
2) A few seconds later, I would be booted out of Github.
3) When I sign in using Github a second, I would remain logged in now.

But then, Github asked me to reauthorize my account, so am wondering if the issue I saw will be gone after this. Also wondering if anyone else is able to reproduce the same issue.

distalx:
@lpatmo That is strange,  I'm not able to reproduce it.

PR #856, opened 2018-04-28
#830 Add global activity feed on home logged out page

### LINKED ISSUE #830: Add some global site numbers to the landing page
On the landing page on codebuddies.org, let's show some global site numbers:

- number of hangouts scheduled
- number of study groups 
- number of learnings
- number of discussions

screen shot 2018-04-17 at 1 23 10 am

railsstudent:
Signed-off-by: Connie Leung <cleung334@gmail.com>

Fixes #830 

Add global activity feeds in home logged out page.
The circles are responsive in different browser width.

selection_028
selection_029
selection_030
selection_031


railsstudent:
@distalx  please comment the visibleHangouts publisher in hangouts/publications.js
Moreover, Is it correct to subscribe allStudyGroups, learnings and allDiscussions to find the count?

distalx:
@railsstudent 

Meteor keeps published data in main memory so if reactivity is not necessary we should avoid using `pub/sub`.

Publication `visibleHangouts` will fetch all the documents with matching criteria and cache those documents in main memory. 

Once the document are fetched Meteor will try to keep them uptodate, So any change occur to these documents (i.e. RSVP, increase of view count) would be reflected to the documents on the main memory and then database after that those effected documents will be pushed back to all the other subscribers. 

In this use case we are only interested in _counts_, not the entire documents. We should use Meteor `methods` instead of `publications`.

Here, We can have a single method which could return an Object with all the required stats.

Very Roughly : 
```
Meteor.methods({ 
  getOverallStats() { 
     return  {
       hangout: Hangouts.find().count(),
       learning: Learnings.find().count(),
       ...

     }
  } 
});
```

railsstudent:
@distalx   Feedback incorporated. Call server method to get global activity count and update reactive variables to render the page

railsstudent:
@distalx Fixed.  Store reactive variables in instance.data and HTML can access the variables directly.

distalx:
@railsstudent, made one minor change instead of using global object `data` for adding state to the template we are storing the template state to the reactive var, This lets us isolate reactivity to a particular template instance. And added `instance` global helper which is just a syntactic sugar.


screencapture-localhost-3000-2018-05-10-13_08_21

lpatmo:
Thanks so much @railsstudent! And @distalx for reviewing! ⭐️⭐️⭐️

I'll borrow this code to put these counts on the hangouts sidebar.

PR #862, opened 2018-05-04
Issue 852: created a global silent hangout page

lpatmo:
Created a global silent hangout page. 

## Non-logged-in user: 
screen shot 2018-05-04 at 6 38 06 pm

## Logged-in user:
screen shot 2018-05-04 at 6 44 18 pm

## After user has joined the hangout:
screen shot 2018-05-04 at 6 44 57 pm


lpatmo:
I updated the design:

screen shot 2018-05-08 at 12 23 50 am

lpatmo:
I also added something to the navbar: 
screen shot 2018-05-08 at 12 25 41 am

1) Do you think the navbar announcing the silent hangout is too much? (Reminds me: we should also highlight `Discussions` as a new feature)
2) Feedback welcome on the title on /coworking ("Global Silent Coworking") and on what the text in the navbar announcement should say.

distalx:
This looks great 👏.

Few suggestions 

1) What do you think if we provide a dropdown on a `coworking` page which could have options like `coworking`, `javascript`, `go-lang` etc? By selecting options user can switch between hangouts.
If we implement this we could have a route like `/h/:room` where room could be dynamic value.

2) Sometimes it takes a while for `jitsi` frame to load, during that can we display a loading screen ?


3) To create a coworking hangout we could use the existing template `hangoutFrame` and avoid the repetition of code. we might have to modify that template a little bit.

very roughly :
```
// coworking.html 
{{> hangoutFrame hroom='coworking' htype=silent}}

//hangout_frame.js

'click .load-hangout': function(event, template){
  const data = {
      room: this._id || template.data.hroom,
      username: Meteor.user().username,
      type: this.type || template.data.htype,
      avatar: Meteor.user().profile.avatar.default
    }
...
```
if we use the `hangoutFrame` template then we don't have to worry about **2**, that case has been handled in `hangoutFrame`.

lpatmo:
@distalx Thank you so much for reviewing! 

> What do you think if we provide a dropdown on a coworking page which could have options like coworking, javascript, go-lang etc? By selecting options user can switch between hangouts.
If we implement this we could have a route like /h/:room where room could be dynamic value.

I think this is a great idea if we have a lot of user traffic, but I'm not sure if this page/this idea will take off yet. Ideally, if this feature were successful, I'm envisioning there would be ~3 people in there at *any* time of the day. So I vote for bumping the idea of having additional rooms to version 2, after 
version 1 gains some traction!

> Sometimes it takes a while for jitsi frame to load, during that can we display a loading screen?
Good idea! You're talking about using the `onload` handler from https://github.com/jitsi/jitsi-meet/blob/master/doc/api.md, right?

> To create a coworking hangout we could use the existing template hangoutFrame and avoid the repetition of code. we might have to modify that template a little bit.
Will look into this! Thanks for the example code.

lpatmo:
I'm going to merge this in now so we have a demo for staging, and I will file a new issue for the `onload` spinner feature.

PR #872, opened 2018-05-13
Enable docker containers for PR's

nalbina:
This will effectively tag pull requests on commit using docker-hub.
I have added a dummy `docker-compose.test.yml` in order to trigger the `post_test` hook.
What this will do is push a docker container to `codebuddiesdotorg/codebuddies:PR_$PRNUMBER` which we can then take from there for demo purposes.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,3 +1,3 @@
 sut:
   build: .
-  command: env
+  command: echo "DONE"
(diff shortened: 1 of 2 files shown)


nalbina:
Let's wait till docker-hub creates a `codebuddiesdotorg/codebuddies:PR_827` ;)

nalbina:
Worth noting is that we hook into the post_test, because we won't get a build or push event for external pull requests (e.g forks)

nalbina:
It works, waiting for your review @distalx @lpatmo

distalx:
Looks good to me.

-----Original Message-----

PR #925, opened 2018-07-28
FontAwesome cleanup: use fas instead of far (pro) icons

lpatmo:
Wrote a small python script to search through the /client directory and find and replace "far fa-" with "fas-fa". 

Background: according to https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#changes, we can only use `fab` and `fas` icons under the free plan. PRO plans cannot be used by open source projects.

### DIFF
--- a/client/css/_hangout_list.scss
+++ b/client/css/_hangout_list.scss
@@ -107,7 +107,7 @@
 
   label {
     input[type="radio"] {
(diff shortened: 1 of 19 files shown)


lpatmo:
@distalx @railsstudent Thank you!!

PR #946, opened 2018-08-18
remove hooks for docker build

sebbel:
This removes the hacky hooks for building PR's on dockerhub which are forks. Let's remove this as it's also proven not to work as intended.


lpatmo:
Thanks @sebbel!

Issue #932, opened 2018-08-02
On codebuddies.org/slack, be clearer about where commercial posts should be posted


lpatmo:
Thanks to a recent PR, we've got a nice, updated version of the Slack etiquette guide published (you can see the PR on staging now: https://staging.codebuddies.org/slack).

However, we don't clearly explain the rules about commercial posts, such as:
- promoting ticket sales to a conference
- promoting an external Slack or Discord or community 
- promoting a book/course one's written that isn't free
- a job posting

Below is a rough draft I wrote up quickly. Feedback welcome on this draft! Let me know if you think a point is off or should be changed, etc.

### Where should I share a commercial post on Slack?

Occasionally you'll want to promote a project you've been working on that asks for money. Should you share it? Absolutely! But please respect the following guidelines about posting commercial content: 

1. In general, all commercial posts can be shared in the `#promote` channel. These include links to ticket sale pages to conferences, requests for members of CodeBuddies to sign up to join an external link, or a link to purchase a course you created that's behind a paywall.

#### Job postings
2. If you want to hire someone or want to forward a job posting, we'd love for you to share the opportunity in #job-postings! Many members keep an eye out there for opportunities.

#### Promoting a paid book or course
3. If you're the author of a course or a book that you think might benefit the community, _congratulations_ on the achievement! Feel free to mention it casually in conversation so that members of the community can ahem, potentially start study groups around the content you've created. Or potentially give you feedback/serve as beta readers/ask you interesting questions. 

If your book/course is behind a paywall and you'd like to sponsor a giveaway that we'll announce to the community (eg 1-2 passes to your book or course), please fill out __this form__. 

#### Promoting conference ticket sales
4. If you're the organizer of a conference that isn't free to attend, but you want to share a link to the ticket sales page in specific channels -- e.g. if you are an organizing a global javascript conference in NYC, and want to share it in both #loc-nyc and #javascript, please find out if you're able to give the CodeBuddies community a discount or a free ticket to raffle away. Then, please fill out __this form__. CFP links don't fall under the same scrutiny and can be posted anywhere/anytime. 

#### Looking for signups
[...]

PR #957, opened 2018-09-06
Fix Issue-932 

peoray:
Fixes issue #932 

The following changes were made:

1. Added information about commercial posting to the slack page
1. Moved the `cb-logo` styling to `base.scss` and removed inline styling
1. Removed unnecessary periods from the page as cited by @wuworkshop 

I would suggest this change is squashed into a single commit maybe as described by the issue created.


lpatmo:
👏 Thank you so much, @peoray!

PR #961, opened 2018-09-11
Fixed: Jarvis constantly returns "Date unrecognized. Please try Again." even for valid date! 

gauravchl:
It was an issue with handling the javascript Object.
Assigning an object from one variable to another doesn't create a new copy, it only passes the object reference. Like a pointer in c.
Solved by destructuring and returning a new copy.


lpatmo:
👏 Thanks so much, @gauravchl.

PR #996, opened 2018-10-05
Add contributor codemzy

### LINKED ISSUE #991: On profile page, add a textarea so users can type in something they learned directly above their log of learnings
If you go to your profile page on https://codebuddies.org, you'll see a log of learnings that you've shared.

screen shot 2018-10-02 at 3 02 15 am

Currently, there is no way to share a learning directly on this page.

Borrowing from the "Today I learned..." section from https://codebuddies.org/hangouts, we should add a textarea here that'll let users share a learning directly.

screen shot 2018-10-02 at 3 02 40 am

## Note to contributors
Note: At first glance, this should be a fairly easy issue. We already have a template called `hangoutLearned`, so I think we can just drop it into the profile page using Spacebar syntax, like so:

```
{{> hangoutLearned }}
```

If you're unfamiliar with this syntax, please read through https://meteorcapture.com/spacebars/. Spacebars is a templating language inspired by Handlebars.

codemzy:
Fixes #991.

Added as a contributor following merge of my pull request #993 for this issue.


lpatmo:
@codemzy Thank you!!

PR #1001, opened 2018-10-07
Home css variable fix

MattKomorcec:
Replace hardcoded value of #fff with a variable $white in _home.scss file.


stain88:
Seems to be a lot more going on in this PR than just a variable change. Why so many files touched?

stain88:
I believe you want to be branching from `staging`, rather than `master`, but pinging @lpatmo @distalx (@railsstudent ) to confirm

distalx:
Yes, please take a pull from the staging on your current branch and then push that changes back here. 
Thanks 

-----Original Message-----

MattKomorcec:
This is such a mess, I'll create a new pull request, that seems like a faster option :-)

Issue #964, opened 2018-09-15
Star (or highlight) active groups on the /studygroups page


lpatmo:
Thanks to @clockwerkz for this idea:

Clicking through the groups, it's hard to figure out which ones are active unless you click through and look through progress logs or past hangouts. 

Would be helpful to highlight the active ones.

lpatmo:
The specific page I'm talking about is this one:
![image](https://user-images.githubusercontent.com/4512699/46131189-e6e65780-c1ef-11e8-9d3f-22f7aa87655e.png)

As users scroll and join groups, it can be hard to tell which study groups are still active. 
(Note: in reality, most activity I've seen from study groups has moved to private channels or DMs on Slack. But let's assume that we want to figure out a way to encourage more usage of this feature.)

Steps for contributors:
1) Read https://docs.codebuddies.org to get set up
2) We have to decide how to define "activity." Last time any data associated with a study group was updated, maybe? If so, we'd need to create a new `updated_at` field for the `Study Groups` collection, and update it with a timestamp whenever new data is saved into a particular study group's document.
3) Maybe we could put a "Last activity: [timestamp]" text on the card for the study group itself.

Thoughts/other ideas welcome on the proposal above.

mahamshahid18:
Hi @lpatmo I'd like to work on this issue.

I like your suggestions. Other things that can be done are:
* Add a `Created On` field to the db and also add it on the card
* As a further addition to this, maybe as an extension for the future, the cards can be filtered by their `created on` date or by latest activity (as you mentioned above)

mahamshahid18:
This is great, this really helps. I'll start on this in a bit ^_^

lpatmo:
Note: we may have to double check that `createdAt` is actually being saved every time a new study group is created. @mahamshahid18 I'm also happy to file this as a separate Github issue in case it is a real bug, and you can continue working on this bug after that issue is fixed.

mahamshahid18:
I'll take a look at this and confirm if it is indeed a bug. I'm working on the hangouts card issue right now. I'll start this one afterwards 👍

mahamshahid18:
@lpatmo just checked, the `createdAt` field is being saved, so we don't need to worry about that 😅

mahamshahid18:
Hey @lpatmo I've added the logic to update the value of the `updatedAt`  field when a study group document is modified. This is working (it's getting updated).
[...]

PR #1002, opened 2018-10-07
Add a button to load more Study Groups

### LINKED ISSUE #938: BUG: Study Groups page doesn't list all groups
Bug

What is the current behavior?
On the Study Groups page (https://codebuddies.org/study-groups), when you scroll all the way down to the bottom of the page it doesn't load the full list of study groups. For example, I don't see the new "Colt Steele's JavaScript Algorithms and Data Structures Masterclass" study group. Searching for it does work though.

You should be able to see the full list of study groups or if there are a lot of study groups, then perhaps some sort of pagination.

I see this in Firefox Quantum 61.0.2 and Chrome 68.0.3440.106 on OS X El Capitan.

AnnaDodson:
Fixes #938.

When lazy loading the scroll listener was triggered multiple times,
which caused the page view limit to increase before more Study Groups
had been loaded in. This set the status flag to true making it think
that there were no more Study Groups to show and didn't allow the user
to load any more in by further scrolling.

While investigating this issue I also discovered that when lazy loading,
the Study Groups would disappear and the loading status would
momentarily flash up. This was caused by `subscriptionsReady` being
false when getting more Study Groups.

Aligned Study Groups page with Hangouts page by replacing
`scrollHandler` with `addMoreStudyGroups`, and styling the load more
button to match.  This also removes a potentially performance intensive
listener to window scroll events.

### DIFF
--- a/client/css/_buttons.scss
+++ b/client/css/_buttons.scss
@@ -272,3 +272,7 @@ li.delete-hangout, li.edit-hangout {
   margin: 0 auto;
   max-width: 340px;
 }
+
+.btn-load-more {
+    margin: 20px auto;
+}
--- a/client/templates/study_groups/all_study_groups.html
+++ b/client/templates/study_groups/all_study_groups.html
@@ -30,62 +30,62 @@
       {{> studyGroupsSearchResult}}
     {{/if}}
 
-
     {{#unless sgSearchMode}}
       <div class="container margin-top-2">
 
-        {{#if Template.subscriptionsReady}}
-          {{#if studyGroups.count}}
+        {{#if studyGroups.count}}
           <div class="row">
             {{#each studyGroups}}
 
               <div class="col-md-4">
-              <div class="well card">
-                <h4>
-                  <a href="{{pathFor 'study group' studyGroupSlug=slug studyGroupId=_id}}">  {{title}} </a>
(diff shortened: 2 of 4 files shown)


lpatmo:
Thank you so much for figuring this out, @AnnaDodson! 

Forgive me if I'm missing something; I see that we're sorting by `start: flag` on line 29 in all_study_groups.js instead of `title: 1`, but I don't think we have a `start` field. (Btw, in case you did not know, a trick to previewing the schema on any local page is to type `ctrl-m` with localhost:3000 running.)

Maybe we should order by `createdAt` instead?

Although... when I created a group locally just now, I noticed that `createdAt` isn't getting saved in the DB. Oops. That might be another bug to file.

AnnaDodson:
Hi! @lpatmo, no problem!

Thanks for spotting that, sorry I missed it! I was finding that the study groups were coming through from the database in order of _id (I think - not 100% clear!) and then being ordered on the front end alphabetically by title so would look like they were 'jumping' around the page on every get more action. I'll re-visit this tonight/tomorrow and update it so they're getting them in the correct order and they're being sorted correctly on the front end in the same way.

Yes, `Ctrl+m` is the best thing! Super impressed by it! 😄

lpatmo:
Thanks @AnnaDodson! Yeah, double checking a potential issue with `createdAt` in #964 makes sense.

PR #1017, opened 2018-10-15
Remove the loading screen when loading more on Hangouts Page

### LINKED ISSUE #1003: Hangouts Load More button takes you to the top of the page after clicking it
Bug. Related to the PR #1002 for issue #938 

What is the current behavior?
Currently, when clicking on the load more button, more Hangouts are loaded but the page scrolls to the top and you need to scroll all the way back down again to find your place.

I would expect to click the button and the Hangouts be dynamically loaded in so my place isn't lost on the screen.

Affected on Firefox and Chrome on Ubuntu and Android Firefox on mobile 

I'm happy to raise a PR for this if required.

AnnaDodson:
Fixes #1003

When the Load More button is pressed, the hangouts disappear and
the loading state momentarily flashes up. This was caused by
`subscriptionsReady` being false when getting more Hangouts.

I also moved the current and past headers into the `if` conditionals to
only display if there are any current or past hangouts to display

### DIFF
--- a/client/templates/hangout/hangout_cards.html
+++ b/client/templates/hangout/hangout_cards.html
@@ -1,48 +1,42 @@
 <template name="hangoutCards">
-  {{#if Template.subscriptionsReady}}
-  <h3 class="hangouts-title">{{_ "upcoming_events"}}</h3>
-    <div id="active_hangouts">
 
-      {{#if liveHangoutsCount }}
-      {{/if}}
-      {{#each hangouts 1}}
-        {{#unless isHangoutCompleted end}}
-        <div class="card">
-          {{> hangoutCard}}
-        </div>
-        {{/unless}}
-      {{/each}}
-
-    </div>
-
-    <h3 class="hangouts-title">{{_ "past_events"}}</h3>
-    <div id="past_hangouts">
-      {{#if pastHangoutsCount }}
-      
-      {{/if}}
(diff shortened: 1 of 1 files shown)


sebbel:
hey @AnnaDodson, thanks for your PR :raised_hands: 

I've deployed your changes to https://codebuddies-1017.now.sh for easier review and also created some bogus hangouts to checkout the Load more button.

The browser doesn't escape to the top anymore! :D 
One thing that I noticed is that I can click the Load More Button one more time when all Documents are loaded, but I am not sure how this was handled in the original implemention.

lpatmo:
@sebbel That "load more button" behavior when all the documents are loaded was a "feature" in the original implementation too. :P Thanks for the PR preview deployment link. 

@AnnaDodson Sorry it took me a while to review your PR, and thanks for this. An aside: have you added yourself as a contributor yet per the instructions at https://docs.codebuddies.org/code-contribution/recognition? If you're interested, feel free to open up another PR for that!

AnnaDodson:
Hi @sebbel and @lpatmo! In reference to your point above "... click the Load More Button one more time when all Documents are loaded..." I noticed this too, I can have a look if you like?

lpatmo:
Thanks @AnnaDodson! I don't think this is a big issue, but if you want to take a look to see how it might be fixed, I won't stop you. :)

Issue #1009, opened 2018-10-11
CSS adjustments for hangout card


lpatmo:
1) When the study group title is too long, it is too close to the "i" icon

2) Expand and center the "RSVP to this Hangout" button; right now it is very narrow.

Let's make the button look more like:

![image](https://user-images.githubusercontent.com/4512699/46783710-288efc00-cce0-11e8-9eb2-75cd0c537aa9.png)

3) Remove the "Join now" green button from "LIVE NOW"/active events

mendaomn:
Hey, I'd like to take up the issue! :) Trying to set everything up right now

lpatmo:
I just realized that all my copy/paste screenshots were off. Will edit the post shortly!

jhorner32:
@mendaomn, let me know if you need any help. I'd love to help on this as well :)

mendaomn:
Hey, I'm working on the issue

How can I reproduce a "live now" hangout? I've gone ahead and set `feeder: true` but it doesn't populate with live events

lpatmo:
@mendaomn You need to click on the "schedule a hangout" button on codebuddies.org/hangouts. The default start time will be "now," so when you schedule one the hangout will be scheduled to start automatically, and will be "live now."

Oh, yes -- the title should be blue, according to the screenshot. Thanks for catching!

PR #1018, opened 2018-10-17
CSS adjustments for hangout card (#1009)

mendaomn:
Hi,

As per #1009 

1. the title doesn't seem to collapse with the "i" icon! 
2. expanded and centered the RSVP button 
3. removed the JOIN NOW button, it's commented as of now. Waiting confirmation that's the right button to remove

Please let me know how I can improve this

### DIFF
--- a/client/css/_hangout_card.scss
+++ b/client/css/_hangout_card.scss
@@ -15,13 +15,19 @@
   }
   border-radius: 0 !important;
   background: $white;
-  display: inline-block;
+  display: inline-flex;
+  flex-direction: column;
+  justify-content: space-between;
   height: 300px;
   width: 100%;
   margin: 5px 0px;
   padding: 15px 20px;
   overflow: hidden;
   position: relative;
+
+  .card-details {
+    flex: 1;
+  }
 
   p.overflowtext {
     color: $secondary-grey;
@@ -31,7 +37,7 @@
     line-height: 30px;
     margin-top: 6px;
     margin-bottom: 10px;
-    @include font-face(600, 20px, normal, $black);
+    @include font-face(600, 20px, normal, $dodgerblue);
     text-align: left;
     text-transform: capitalize;
     &.completed {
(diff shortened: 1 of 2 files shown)


sebbel:
Hey @mendaomn thanks for your PR.
I've deployed your changes to https://codebuddies-1018.now.sh/ for easier review and created some test hangouts.
Really like the modern look you gave those hangouts! :heart:

mendaomn:
> @sebbel Wish login could be figured out, but even without it that's a brilliant way to preview PRs! Thank you! :)
> 
> @mendaomn Yes, that is the correct button to comment out. Your code changes look good! Only issue I noticed: when the hangout title is three lines long (see the first card on https://codebuddies-1018.now.sh/hangouts), the content of the card shifts/moves. Let me know if you don't see an easy fix for that. Thanks.

I tried to work around that for a while, still haven't come up with a solution but it shouldn't be too hard to pin point to its cause. I'll dig deeper!

mendaomn:
> @sebbel Wish login could be figured out, but even without it that's a brilliant way to preview PRs! Thank you! :)
> 
> @mendaomn Yes, that is the correct button to comment out. Your code changes look good! Only issue I noticed: when the hangout title is three lines long (see the first card on https://codebuddies-1018.now.sh/hangouts), the content of the card shifts/moves. Let me know if you don't see an easy fix for that. Thanks.

I tried to work around that for a while, still haven't come up with a solution but it shouldn't be too hard to pin point to its cause. I'll dig deeper!

mendaomn:
> Let me know if you don't see an easy fix for that. Thanks.

I tried to work around that for a while, still haven't come up with a solution but it shouldn't be too hard to pin point to its cause. I'll dig deeper!

lpatmo:
@mendaomn My guess is has something to do with the flex property, but perhaps one workaround could be to remove the "Scheduled by [username] with # RSVPed" from all the hangouts that are upcoming. 

I tested this out on the browser and it looks okay; let me know if this makes sense to you too!

mendaomn:
Sounds good to me!

lpatmo:
Hi @mendaomn! Any chance you'd be able to make a commit removing the "Scheduled by [username] with # RSVPed" this weekend? Otherwise, no worries, I will make the change. Thanks for your time!

lpatmo:
(Actually, going to go ahead and merge this in. Will file a new Github issue for the edge case, which you are free to work on if you comment on it. Thanks!)

PR #1025, opened 2018-10-20
Updated SweetAlert2 with meteor compatible sweetAlert2

### LINKED ISSUE #1014: Upgrade sweetalert 1.x to sweetalert 2.x
I am filing this bug after discussion in https://github.com/codebuddies/codebuddies/issues/821 (adding a "sign in using Github" link on the sweetalert) surfaced the fact that we need to upgrade to sweetalert 2.x before we can support having both a Github and Slack sign-in actions on the sweetalert.

The task:
- upgrade to sweetalert 2.x from 1.x
- Guidelines to follow: https://sweetalert.js.org/guides/#upgrading-from-1x

thunatos:
Fixes #1014
sweetalert2 was not properly being used, package has been updated to support Meteor sweetalert2. 

Prior to change, code as such was unable to display properly:

```
swal({
  title: '<strong>HTML <u>example</u></strong>',
  type: 'info',
  html:
    'You can use <b>bold text</b>, ' +
    '<a href="//github.com">links</a> ' +
    'and other HTML tags',
  showCloseButton: true,
  showCancelButton: true,
  focusConfirm: false,
  confirmButtonText:
    '<i class="fa fa-thumbs-up"></i> Great!',
  confirmButtonAriaLabel: 'Thumbs up, great!',
  cancelButtonText:
    '<i class="fa fa-thumbs-down"></i>',
  cancelButtonAriaLabel: 'Thumbs down',
})
```

Where html code was not being bound properly. With the updated change, you can now see html tags being used fully inside of a sweetalert.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -10,10 +10,8 @@ mongo@1.5.0                   # The database Meteor supports right now
 blaze-html-templates    # Compile .html files into Meteor Blaze views
 session@1.1.7                 # Client-side reactive dictionary for your app
 tracker@1.2.0                 # Meteor's client-side reactive programming library
-
 es5-shim@4.8.0                # ECMAScript 5 compatibility for older browsers.
 ecmascript@0.11.1              # Enable ECMAScript2015+ syntax in app code
-
 accounts-password@1.5.1
 fourseven:scss@4.9.0
 check@1.3.1
@@ -71,3 +69,4 @@ underscore
 twbs:bootstrap@=3.3.5
 percolate:migrations
 
+cleandersonlobo:sweetalert2
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -26,6 +26,7 @@ caching-html-compiler@1.1.3
 callback-hook@1.1.0
 cfs:http-methods@0.0.32
 check@1.3.1
+cleandersonlobo:sweetalert2@1.5.0
 coffeescript@1.0.17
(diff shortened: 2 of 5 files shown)


lpatmo:
@thunatos Thanks for this PR, and for finding that new sweetalert2 package! Just occurred to me that the upgrade *may* break some of the places where we have sweetalert 1.0 syntax, but have not checked. Will do later and see if I have to file a new issue to properly upgrade all those swals. :)

lpatmo:
@thunatos This PR looks good, but I am going to hold off from merging it in to staging (staging.codebuddies.org) just in case it breaks a lot of the swals, and it'll take longer than we expected to merge all of them in. If you get a chance to take a look at which swals it seems to be breaking and would like to open up another PR to fix that, that would be fantastic.

thunatos:
@lpatmo I figured out what the reasoning was and why most things are broken after upgrading to sweetAlert 2. 

Refer to this stack overflow post: https://stackoverflow.com/questions/41460989/meteor-call-dont-work-inside-sweetalert-confirmation-box

So I will begin updating the code so that it will be adjusted to swal2 promises instead of callbacks. 

👍

lpatmo:
@thunatos Thank you! I filed an issue for this: https://github.com/codebuddies/codebuddies/issues/1034

Please comment in the thread if you get stuck/if parts of it aren't very straightforward.

PR #1048, opened 2018-10-27
Update print width from default for configs

### LINKED ISSUE #1045: increase prettier's default line length to 100.
default line length is 80, increase that to 100. [here](https://github.com/codebuddies/codebuddies/blob/staging/package.json#L17)
i.e. `"prettier --print-width 80 --write"`
Ref : https://prettier.io/docs/en/options.html#print-width

JackHowa:
Fixes #1045 

Briefly saw that the length of prettier line breaks need to be longer than default. Looked up and then tested the command. 

referenced https://prettier.io/docs/en/options.html#print-width

screen shot 2018-10-27 at 2 41 58 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "lint-staged": {
     "*.{js,json,css}": [
-      "prettier --write",
+      "prettier --write --print-width 100",
       "git add"
     ]
   },


lpatmo:
Looks good; thanks, @JackHowa!

PR #1067, opened 2018-10-29
Fix slack hangout reminder alert issue

### LINKED ISSUE #1020: Slack reminder alerts are not sending out
We have a bit of code that sends out reminders to the #announcements channel that a hangout is set to start in less than 2 hours (see: `server/slack/methods.js`): 

```
if (type === "REMINDER") {
    fallback =
      `Reminder! hangout: ${hangout.topic} is set to start on ${
        hangout.start_time
      }. Visit` + Meteor.absoluteUrl("hangout/" + hangout._id);
    pretext = `Reminder! ${hangout.type} hangout: *${
      hangout.topic
    }* scheduled by a CodeBuddies member is set to start in *less than two hours!* ${nameString}`;
  }
```

However, although announcements about new hangouts are firing, this reminder code is not. Wonder if someone could look into it.

yoksanherlie:
Fixes #1020 

Actually, the bug rise from the error of the html template for the e-mail notification. The notification works well.

### DIFF
--- a/private/email-hangout-reminder-two.html
+++ b/private/email-hangout-reminder-two.html
@@ -13,7 +13,7 @@
           <tr> 
               <td>
 			     <p style="margin: 0px;">Hi there!</p>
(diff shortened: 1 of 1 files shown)


lpatmo:
Ah, so the Slack reminder alerts were working well on #codebuddies-ops? Thanks for spotting this HTML issue in the templates!

Issue #913, opened 2018-06-16
feat-inbox


distalx:
Problem:
Since now we have added a signin with GH added to the CB it’s a bit difficult to get in touch with any one who joins a studygroup or hangout after logging in with GH.

This will feat will provide a way to directly communicate with other members on the platform.

Specs:
User can start a new conversation by selecting a user by their username.

It will create a new private thread between the sender and the recipient.

Sender and recipient would be able to respond to the thread 

Questions:
Would the inbox be a page (like messenger.com), or a popup a user could open up from any page?
How many messages should be displayed at once? 20?
Notifications about new “inbox” messages should be separate from our existing “notifications” tab, probably

Future features:
Hangout organizer should be able to send a message to the inboxes of everyone who RSVPed to a hangout (This is a priority feature IMO. And I am tempted to have those messages go into people’s emails directly, although emails should probably be opt-in. So at the very least, the messages should go into an inbox.)
Checkbox to subscribe to email notifications from inbox messages 
Group thread (3+ people in the same inbox)

sergeant-q:
@distalx a couple of questions and comments. First off, I really like this idea for a few points:

1 - An opportunity to move discussions (and traffic) more onto the platform itself
2 - Increase overall engagement between the community members
3 - Improve the knowledge base and documentation of learnings

However, before we truly consider this "inbox" feature. I am going to pose some questions that I'd like everyone to consider and brainstorm about:

 - Prior to GH login being an option, how did hangout organizers and participants communicate with each other?
 - Would creating chat capabilities within the platform be a better approach as opposed an "inbox"? Or what are the benefits of an inbox thread over chat?
 
As much as I like the potential of this feature moving more traffic to the platform, with communication tools like Slack already in place - are we re-inventing the wheel here?

Just some food for thought. As we build out the CB site and features, I want to keep in mind our purpose and goal. We certainly want to build a self-sustaining platform and some features will help do just that. However, we also need to ensure we are leveraging existing solutions if it makes sense.

Thoughts?

nalbina:
basically reinforcing what @sergeant-q said here. Offloading communication to Slack is totally fine.

sergeant-q:
This also pushed the need to revive the forum idea. I think the self documenting knowledge base for our community can only be possible this way. I think study group discussions can be effectively managed via private discussions on the forum.

lpatmo:
> Prior to GH login being an option, how did hangout organizers and participants communicate with each other?

[...]

PR #1076, opened 2018-11-12
Feat inbox

distalx:
ref #913.

This is an experimental feature, it will provide a private form of communication between users that is only visible to the sender and recipient.

- User can start a new conversation thread by clicking on **message** button, from a profile page. 
screen shot 2018-11-12 at 8 42 41 am

-  it will create a new thread.
screen shot 2018-11-12 at 9 30 03 am

user will receive an email notification for unread messages. user can set their email notification preference form settings.


distalx:
Merging this to the staging branch so can test out on `staging.codebuddies.com`, but the PR still needs a code review.

Also, planning on adding/improving following things into a next iteration.
- Send button for mobile device.
- UI for mobile device.
- Support for Emojis.
- Search option!

Issue #958, opened 2018-09-08
Future @cb_jarvis upgrades


lpatmo:
(Thanks @julianjohannesen for feedback)

- pass in description
- pass in hangout type (silent, collab, teaching)
- pass in external link in case folks want to use Zoom, LiveStream, etc.

gauravchl:
More command ideas for jarvis:

1. `@cb_jarvis stats`
    o/p: Total hangouts organized, learnings reported, study groups, discussions, online users, total users etc...
2. `@cb_jarvis 24/7 hangouts`
    o/p: all 24/7 hangout detail(title/studyGroupName) with jitsi link
3. `@cb_jarvis #TodayILearned`
   o/p: Random "today i learned" post from last 10 posts. With user name.

lpatmo:
Love all those ideas. If they can be pulled into separate PRs/issues, we can definitely do that so that you (and potentially other contributors!) get max credit. :)

gauravchl:
Good idea. So let's create new issues for every command with `#help wanted` and `#good first issue`?
https://help.github.com/articles/helping-new-contributors-find-your-project-with-labels/

Does [theses projects](https://github.com/codebuddies/codebuddies/projects) are still relevant? 
May be we could keep above tasks well organized there as well.

lpatmo:
> Good idea. So let's create new issues for every command with #help wanted and #good first issue?

Sounds good! I will start doing that. 

> Does theses projects are still relevant?

Er... wow, I totally forgot about those projects. Thanks for reminding me. I think some of those issues are relevant still, but working with projects hasn't been very intuitive, so I will try to turn them into issues. I've mostly still been using the Trello board at https://trello.com/b/3Olzq5sh/codebuddies.

That said, if you like the Github Projects feature, I don't trying it out! I'll create a new board called `@cb_jarvis: Slack to CodeBuddies` or something.

railsstudent:
@lpatmo @gauravchl  Should create hangout command split into create silent hangout, create collab hangout and create teaching hangout commands respectively or add a new segment to create hangout to hold hangout type?

lpatmo:
Sadly, not many people are using the `@cb_jarvis` command to create new hangouts. Besides `hangout type`, there are two other arguments we could add:
- specify the description
- specify an optional external link (like Twitch or Zoom) - requested by @julianjohannesen 

[...]

PR #1084, opened 2018-12-14
Add hash tag support for jarvis

gauravchl:
- HashTag support added
- Code refactored into modules(Commands, Actions, Helpers and WebHooks)

No more `@cb_jarvis` mentioning. Simply use hash tagging!
Supported Hashtags:

```
#createHangout tomorrow at 6 pm, YOUR HANGOUT TITLE
#til #Django crispy forms rocks! 
#listHangouts
#howAreYou?
#help
#help Who is strongest avenger?
#til
```
All commands are case insensitive.

screen shot 2018-12-14 at 1 47 14 pm

Issue Covered: https://github.com/codebuddies/codebuddies/issues/958


d3vild06:
Hope to get this reviewed by this weekend

lpatmo:
Thank you so much, @gauravchl 🙏 Didn't expect a PR so quickly; we may get a release out this weekend w/ inboxes. 

One thought. Do y'all think we should support: 

`#hangout in 1 hour, Advent of Code Day 5`?

The command is more vague, but it's less typing. In any case, I really hope we get in the habit of using it. :)

billglover:
I’m against hashtags for two reasons:
- they autocomplete as channel names
- Slack has a standard interface for commands, the `/command`

lpatmo:
Whoops, I did not think very seriously about possible interference with channel names. Where were you four days ago in https://github.com/codebuddies/codebuddies/issues/958#issuecomment-446081133, Bill? (JUST KIDDING! I really should've done another quick user poll in #codebuddies-meta)

We rarely use the `/` command in Slack (totally forgot about it), but I can see something like:
`/hangout tomorrow at 9pm, Advent of Code` or
`/til message`
... being a standard.

What do you think, @gauravchl? Would it be as straightforward as the change below, plus all the additional hashtag replacements? Or do you prefer hashtags over using the slash command?

```
  TIL: ["/til", "til", "/todayilearn"],
  HELP: ["/help"],
  CREATE_HANGOUT: ["/hangout"],
  LIST_HANGOUTS: ["/listhangouts", "/listhangout"],
  HOW_ARE_YOU: ["/howareyou", "/howareyou?"]
```

billglover:
Apologies, not been online much. I'm not so sure it is a straightforward change (different Slack API endpoint) so it may be worth releasing and requesting feedback from users.

We definitely don't make use of Slack commands, but they are fairly common and new users may expect that is where they find things they can do. Might be worth a broader discussion (under a separate issue) before changing anything.

gauravchl:
@lpatmo @billglover 

Yeah, i do agree with Bill. Fair points.
Slash command is actually built for performing such actions.

I think the main question would be how other users are familiar with those commands. Hash tags are very popular and is one of the reason why we are replacing `@cb_jarvis` with hash.

Personally i don't mind using slash or hashtaging or even `@cb_jarvis`.
But yes i would prefer the standard way(slash commands).

Proper way to implement slash command is a bit different than current implementation, might need to setup different endpoints and webhook.

gauravchl:
screen shot 2018-12-16 at 1 41 36 pm

gauravchl:
@lpatmo @billglover Slash command support added for all jarvis commands https://github.com/codebuddies/codebuddies/pull/1084/commits/3f769de4bf76596b6e19064d8cf1be1c098c509b

So for now it will support both hashtag and slash commands.
For slash commands process is very simple-
1. Go to slack app settings and open `slash command` tab: https://api.slack.com/apps/ACAB1TCTD/slash-commands?saved=1
2. Create new commands there, make sure command name is be same as listed here https://github.com/codebuddies/codebuddies/blob/51df2fd0f4f843116802ee0a5aeb070169501eb9/server/slack/commands.js#L2
3. Request url should be in form `https://6c0634ca.ngrok.io/webhook/slack-command`
 or `https://codebuddies.org/webhook/slack-command`
4. Save it. Reinstall the app to slack if required(first time).

lpatmo:
@gauravchl Thank you for the additional work! I have added the slash commands. 
image

server/slack/commands.js
```diff
@@ -0,0 +1,10 @@
+// All commands are case insensitive
+const Commands = {
+  TIL: ["#til", "til", "#todayilearn", "/til", "/todayilearn"],
```
lpatmo: `#todayilearned` instead of `#todayilearn`, maybe? (Will update the corresponding command in the Slack app if you make this change.)

PR #1108, opened 2019-01-26
Sort study groups on codebuddies.org/study-groups by "most recent activity"

### LINKED ISSUE #1107: Study groups page should be sorted by last activity, at least as an option
https://codebuddies.org/study-groups is currently not sorted by last active date, but maybe should be, or that should at least be an option. Not sure what the current sort order and/or if there is one.

lpatmo:
Fixes #1107.


lpatmo:
Thanks @d3vild06! <3

Issue #1110, opened 2019-02-14
IDEA: Feed scheduled hangouts automatically into a google calendar


lpatmo:
Proposed by @gauravchl:

```Feature Proposal: Create a Codebuddies google calendar and whenever a new hangout is scheduled add that hangout entry to calendar as well. Others can subscribe to the calendar and can see upcoming or ongoing hangouts right into their own choice of app(apple/yahoo/google whatever calendar they use).```

lpatmo:
Maybe we should embed the Google Calendar to the sidebar on codebuddies.org/hangouts, if this happens.

gauravchl:
Totally 💯

railsstudent:
Does google calendar required to be under codebuddies gmail account and grant public for codebuddies members to see?

lpatmo:
So in the early days (2014) before CB had a proper events website I actually did use Google Calendar to keep track of hangouts in the Google Groups; here is the public calendar:

https://calendar.google.com/calendar/embed?src=codebuddiesdotorg%40gmail.com&ctz=America%2FLos_Angeles

And embed code: <iframe src="https://calendar.google.com/calendar/embed?src=codebuddiesdotorg%40gmail.com&ctz=America%2FLos_Angeles" style="border: 0" width="800" height="600" frameborder="0" scrolling="no"></iframe>

Public address in iCal format:
https://calendar.google.com/calendar/ical/codebuddiesdotorg%40gmail.com/public/basic.ics

lpatmo:
Mockup in browser:

![image](https://user-images.githubusercontent.com/4512699/52778319-9767c300-2ffa-11e9-96ff-acc644aee0a0.png)

gauravchl:
Google calendar will be under Codebuddies Google account. Codebuddies app(Meteor) will use API tokens to add events to it whenever a new hangout is scheduled(+ other CRUD operation). 
https://developers.google.com/calendar/quickstart/nodejs

For other peoples(codebuddies user +non user) this calendar will be read only. and they can subscribe to it using ical feed. Like this one(Liverpoo FC matches): https://calendar.google.com/calendar/ical/p520al5mfgqq5m2a8pu021nv0c%40group.calendar.google.com/public/basic.ics

railsstudent:
> Google calendar will be under Codebuddies Google account. Codebuddies app(Meteor) will use API tokens to add events to it whenever a new hangout is scheduled(+ other CRUD operation).
[...]

PR #1111, opened 2019-02-14
Allow users to list hangouts on their own choice of calendars.

gauravchl:
User can now subscribe to codebuddies hangout using iCalFeed.
This feed url will publish all the upcoming hangouts and hangouts from past month.

**Feed url will be**: https://codebuddies.org/i-cal-feed 

**How to test(screencast)**: https://youtu.be/bD10mGCIo7g

**Note**: Different calendars(Google/Apple/outlook) have different refresh frequency, so if you are already subscribed to feed and then creates a new hangout, it may take some time to appear new hangout. Totally depends upon the calendar's refresh frequency, You can change it on Apple calendar but Google doesn't have this option afaik.

Issue: https://github.com/codebuddies/codebuddies/issues/1110

### DIFF
--- a/server/i-cal-feed.js
+++ b/server/i-cal-feed.js
@@ -1,0 +1,47 @@
+import moment from "moment";
+import ics from "ics";
+
+const iCalFeed = {
+  handleRequest(params, req, res) {
+    console.log("iCalFeed.handleRequest called");
+
+    const hangouts = Hangouts.find({
+      end: {
+        $gte: moment()
+          .subtract(1, "months")
+          .toDate()
+      }
+    }).fetch();
+    const events = hangouts.map(hangout => {
+      return {
+        uid: hangout._id,
+        title: hangout.topic,
+        description: hangout.description + `\n${hangout.externalURL}`,
+        start: moment
+          .utc(hangout.start)
+          .format("YYYY-M-D-H-m")
+          .split("-"),
+        end: moment
+          .utc(hangout.start)
+          .format("YYYY-M-D-H-m")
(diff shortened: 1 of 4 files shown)


server/i-cal-feed.js
```diff
@@ -0,0 +1,52 @@
...
+import ics from "ics";
+
+const iCalFeed = {
+  handleRequest(params, req, res) {
+    console.log("iCalFeed.handleRequest called");
+
+    const hangouts = Hangouts.find({
+      end: {
+        $gte: moment()
+          .subtract(1, "months")
```
d3vild06: @gauravchl can we make this configurable via the meteor settings or env var so we can easily adjust if we choose to in the future?

lpatmo: @d3vild06 Can you elaborate on what we should put into meteor settings or env variables to more easily adjust, Roberto? I'm a little confused, because we don't have any API keys here; going to codebuddies.org/i-cal-feed will download the calendar!

d3vild06: I was referring to the month interval. Instead of hard coding "month" or "1" can we use meteor settings to set them instead.

lpatmo: Got it! Do you think we'll need to change that setting often? I personally don't care too much to see hangouts that happened over a month ago, although it might be interesting to toggle back. 

Although... Meteor settings are a pain to manage since I have to manually add it to a file in the droplet and do a bunch of server restarts, so if we can get away without not putting it in there, then I'm fine with that. :)

PR #1130, opened 2019-03-08
Issue 1122

### LINKED ISSUE #1122: Feature: support text formatting options for posts on the Discussion page
Feature

What is the current behavior?
This would be an expansion of the options for setting up a proposal in "Discussions".
screenshot from 2019-02-25 21-19-03

When I got to 'write' the new Discussion, I would like a tray of options (in the form of icons/text) below the dialog box, as seen in this 1st screenshot:
(disclaimer: I'm using this as an example, directly from my google gmail, I do not own any of the material shown).
screenshot from 2019-02-25 21-23-06

In particular, to create custom html links to reduce the verbosity of posts. 

- It might be good to offer ways to link images and other materials. Attaching docs would be convenient.

- Since we have good google integration, maybe a way to directly share some content from Drive. This would be awesome for shared study materials--could be possible for other people to add materials in the proceeding discussion-thread?

- A way to transfer the discussion attachments and formatting into the subsequent CB-Hangouts I would schedule.

Thanks again!

If there was a keyboard combination that could reach the same effect, this would also work (for me*), as long as there was some way to offer a pop-up or dialogue box that could elaborate.

distalx:
Fixes #1122.

`Styling with Markdown is supported.` added as a help text for better user experience.

screen shot 2019-03-08 at 9 02 59 am


lpatmo:
Thanks @distalx! A suggestion:

image

distalx:
@lpatmo : Some correction - I've only added following text "Styling with Markdown is supported."  [here](https://github.com/codebuddies/codebuddies/pull/1130/files#diff-b88f6b423357879f9fd5ba1ceebce7ddR35). Not modifing the placeholder of the textarea. I accedently uploded wrong image. This is what it will like.
Screen Shot 2019-03-10 at 7 46 03 AM

lpatmo:
Ahh, got it now. Thank you!

PR #1132, opened 2019-03-10
remove column names and add created by to discussion item

### LINKED ISSUE #1120: Issue: Column headings for 'Topic' and 'Author' out of place on Discussions page
On the discussion page, in mobile view, the topic and author text is kind of just lingering there and doesn't look quite right.

image

cbelsole:
Fixes #1120.

As discussed in the ticket I have removed the column name and added the author to the discussion item.

Desktop:
disscussion_full

Mobile:
disscussion_mobile

### DIFF
--- a/client/templates/discussion/all_discussions.html
+++ b/client/templates/discussion/all_discussions.html
@@ -40,13 +40,7 @@
       {{#if Template.subscriptionsReady}}
         {{#if discussions.count}}
           <div class="row">
-            <div class="col-md-8">
-              <h4>Topic</h4>
-            </div>
-            <div class="col-md-2">
-              <h4>Author</h4>
-            </div>
-            <div class="col-md-2">
+            <div class="col-md-12">
               <div class="dropdown pull-right">
                 <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
                     <span><i class="fas fa-filter" aria-hidden="true"></i></span>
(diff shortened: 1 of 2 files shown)


angelocordon:
Looks good to me @cbelsole - will leave it to @distalx or @lpatmo to merge in case they see something I don't - been a while since I touched the codebase.

distalx:
Can we remove the author field altogether from this page?

billglover:
I can see an aesthetic case for removing it,  but does that detract from the site functionality? Is there another reason for removing it?

Retaining it brings the following:
* associates the discussion with an author
* credits the author with contribution to the site
* fits with information people expect to see on a post
* discourages the behaviours associated with anonymity

When browsing discussion topics, I look for two things:
* topics I have an active interest in (by any author)
* topics that I’m unfamiliar with (by authors I find inspirational/credible)

Removing the author field prevents me from doing the latter.

lpatmo:
Thanks for tackling this, @cbelsole! :) 

Originally when I glanced at the screenshot from my phone I was thinking that anonymity would be _good_, because we'd want people to click through to the discussion threads, and also it'd help "protect" shyer posters. Then I got back home and looked at it properly on a checkout, and changed my mind. (And the several reasons Bill listed for keeping the author are valid, too.)

The only thing is the "Created At" all caps and font size discrepancy seemed a bit weird, so I tweaked it slightly so the "author" link would look more uniform:

image

ArcTanSusan:
unrelatedly, what is the "Posted on :" timestamp syntax? Is this by design?

lpatmo:
@ArcTanSusan You're right, that `:` is extraneous. Thank you for catching!

Issue #1098, opened 2019-01-01
STUDY GROUPS: card css could have more height


stain88:
Do you want to request a feature or report a bug?
**Bug**

What is the current behavior?
<img width="384" alt="low_height" src="https://user-images.githubusercontent.com/8461452/50575594-2e850c80-0df9-11e9-9819-4a05ed1f2c6c.png">

What is the expected behavior?
Bit more height to avoid the overflow on the bottom.

Which browser / OS are affected by this issue? (An *isolated* way to reproduce the bug)
FF/Safari/Chrome on Mac, so I'm guessing most.

stain88:
As a second tweak, I'd add a `white-space: nowrap;` on the `small` around the user class, to avoid that breaking on two lines:
<img width="377" alt="split_user_count" src="https://user-images.githubusercontent.com/8461452/50575647-d8649900-0df9-11e9-8eec-8da5b47fd8f3.png">

lpatmo:
I don't see a PR related to this but I _think_ it looks fixed on codebuddies.org/groups now (and maybe I forgot to close). Feel free to re-open if I'm wrong!

stain88:
Both issues are still present on all Mac browsers:

<img width="1325" alt="card issues" src="https://user-images.githubusercontent.com/8461452/52901652-23960800-31fe-11e9-9357-37c86dbcbe0d.png">

lpatmo:
D'oh, you're right, not sure how I confused it with some other cards. Sorry!

stain88:
So, "Large" windows are fine, "Small" windows are fine, but the 992-1200px range is still not quite there. "Readings in Computer Science and Technology" seems to be a particularly long one. But at exactly 992px width, this happens:

<img width="1104" alt="992px_cards" src="https://user-images.githubusercontent.com/8461452/52911656-ab335380-329e-11e9-88d2-8c1788303424.png">

lpatmo:
Ah wow, 992px exactly! I had to install a browser width chrome extension to capture that since I kept sliding the browser past 992px.

akosasante:
I'd like to take a look at this issue if it's still occurring 👋

stain88:
Yep, I'm still reproducing this at exactly 992px width, on several browsers.
All yours, @akosasante

akosasante:
I've opened a PR 😄  I may have mishandled my git-fu though, it looks like it brought in a bunch of extraneous commits :/

[...]

PR #1143, opened 2019-10-11
Resolving Issue-1098

akosasante:
Fixes Issue #1098 .

Bootstrap's `col-md-4` on the card divs on the study group page causes each card to get a width of ~33% on screens greater than 992px wide. There is a SCSS rule that manually sets the height of the cards, but this only takes effect at 993px and up. So there's a 1px range at 992px, where there is no set height, so the cards' heights vary dependng on how much content is inside. I fixed this by changing the media query in the study-groups scss file.

However, this results in some potential UI issues if the content of the card (either the title of the study group or the tagline/description) is too long. Fixed this by forcing the title to only be 1 line of text, with an ellipses if it overflows + setting a max height for the tag line, with a scroll bar if needed. 

Let me know if any further changes are needed or if one of the above proposed resolutions is insufficient. 😄 

Before: 
Screen Shot 2019-10-11 at 3 49 56 PM

After:
Screen Shot 2019-10-11 at 3 49 56 PM

### DIFF
--- a/client/css/_study_groups.scss
+++ b/client/css/_study_groups.scss
@@ -42,7 +42,16 @@
     color: $danger-red;
   }
   .card {
-    @media screen and (min-width: 992px) { height: 190px; padding: 19px 19px 10px; }
+    @media screen and (min-width: 992px) {
+      height: 190px;
+      padding: 19px 19px 10px;
+      h4 a {
+        display: block;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+    }
     @include breakpoint(large) { height: 190px; padding: 19px 19px 10px; }
   }
   p {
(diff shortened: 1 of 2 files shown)


lpatmo:
Thanks so much for taking a look, @akosasante! Two quick thoughts:
1/ At first glance, the two screenshots look the same!

2/ Can you open up a new PR merging your PR against the `staging` branch on CodeBuddies?

akosasante:
🤦🏾‍♀️Aw man @lpatmo , haha yep I uploaded the same screen shot twice; will fix! For #2, can I just change the base branch of this PR to staging? Or is it preferable to just have a clean slate with a new PR?

angelocordon:
@akosasante that should work but should also try to clean up the stray commits 👍

akosasante:
Ah I see now that it's because I used the same branch name; sorry did not realize that initially. Closing this PR in favour of: https://github.com/codebuddies/codebuddies/pull/1144

PR #1172, opened 2021-10-07
Github login debugging

### LINKED ISSUE #1171: Not able to sign up on codebuddies website
Before you open an issue, please check if a similar issue already exists or has been closed before. Also please remember that you can also get help in the [#codebuddies-meta channel on the CodeBuddies Slack](https://codebuddiesmeet.slack.com/messages/codebuddies-meta/). (If you are not a member, you can get your invite from [here](http://codebuddiesmeet.herokuapp.com/).

What is the current behavior?

lpatmo:
Fixes #1171 (hopefully).

This PR follows the guidelines in https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/ to (hopefully) fix the deprecated API authentication method.

Before: 
`curl "https://api.github.com/user/repos?access_token=my_access_token"`

After: 
Send the token in the header like: 

```
curl -H 'Authorization: token my_access_token' https://api.github.com/user/repos
```

It looked like we previously were using the clientId and clientSecret as a query param, so I kept the `auth: `${clientId}:${clientSecret}` and updated the  HTTP call to be `https://api.github.com/user/repos`

image

Meteor docs on HTTP: https://docs.meteor.com/api/http.html

Note: this was difficult to debug in localhost:3000, so I'm making the change against `master`. Should be low risk since only the GitHub API call file is affected, and it's already broken.

### DIFF
--- a/server/github_api_calls.js
+++ b/server/github_api_calls.js
@@ -4,15 +4,17 @@ githubAuthRevoke = function(accessToken) {
   const clientSecret = Meteor.settings.github_clientsecret;
   const options = {
     headers: {
-      "User-Agent": "CodeBuddies"
+      "User-Agent": "CodeBuddies",
+      "Authorization": `token ${accessToken}`
     },
     auth: `${clientId}:${clientSecret}`
   };
   const response = HTTP.del(
-    `https://api.github.com/applications/${clientId}/grants/${accessToken}`,
+    "https://api.github.com/user/repos",
     options
   );
   const { statusCode = null } = response || {};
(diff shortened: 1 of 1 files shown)


lpatmo:
Unfortunately this PR didn't fix login; I'm going to debug later by making some curl calls with the actual clientID / secretID (in case some keys need to be refreshed). Otherwise, I plan to put a try/catch around the call here: https://github.com/codebuddies/codebuddies/commit/20ad4f72b78722ac9207e14ac2d79e53790b98c0

PR #1177, opened 2021-11-22
fix - sign in with GH

distalx:
Production release.

This fixes sign in with GH issue.


client/templates/admin/inc/sys_header.html
```diff
@@ -45,7 +45,7 @@
               </li>
             {{/if}}
             {{#if isInRole 'admin' 'CB'}}
-              {{> loginButtons align="right"}}
+              <!-- {{> loginButtons align="right"}} -->
```
lpatmo, line 48: @distalx Quick question -- why did you comment out these login buttons for admin CB users?

server/github_api_calls.js
```diff
@@ -4,17 +4,17 @@ githubAuthRevoke = function(accessToken) {
...
+      access_token: accessToken
+    }
   };
-  const response = HTTP.del(
-    "https://api.github.com/user/repos",
-    options
-  );
+  // please see the following documentation
+  // https://docs.github.com/en/rest/reference/apps#delete-an-app-authorization--code-samples
+  const response = HTTP.del('https://api.github.com/applications/${clientId}/grant', options);
```
lpatmo, line 15: My bad for the former change here... btw restoring it back to:

image

... should also work, right?

PR #1181, opened 2022-03-25
Releasing the material icons fix

lpatmo:
Merging in https://github.com/codebuddies/codebuddies/pull/1179 from staging into production.


lpatmo:
Note: this won't solve login (which is only working on localhost:3000 but not in staging or prod, but at least will remove the Material errors in prod)

